PR #8, opened 2025-09-15
Remove raw usages of cudaSetDevice, and fix sync on fingerprint tests

scal444:
This fixes a possible error where ETKDG or MMFF did not reset the device after dispatching streams on the initial thread. This could lead to downstream errors in fingerprinting or TS where data allocations and memcpy options were targeting incorrect devices. Would only happen on the same process, such as pytest over the codebase.

The sync issue is separate


evasnow1992:
Looks good to me. I've reviewed the code changes and can confirm the fix addresses the issue correctly. Thank you for the fix!

evasnow1992 (approved):

PR #12, opened 2025-09-22
Remove Bulk Similarity and BitVector-based APIs

scal444:
This PR simplifies our similarity API to only support cross Tanimoto / Cosine on GPU resident data.

The other APIs were inefficient and cluttered the codebase. The bulk case is handled as a subcase of cross with dim 1 for one of the input arrays. The cross APIs with BitVector inputs can be handled in user-space, added examples in docs on how to deal with that.

For the functionality we do keep, improved coverage of unit tests.


nvmolkit/similarity.py
```diff
@@ -29,37 +29,16 @@
...
-                                                             fingerprint_group_two.__cuda_array_interface__))
-    raise ValueError("Unsupported type for fingerprint_one")
-
-
 def crossTanimotoSimilarity(fingerprint_group_one: AsyncGpuResult | torch.Tensor,
                             fingerprint_group_two: AsyncGpuResult | torch.Tensor | None = None) -> AsyncGpuResult:
     """Returns the Tanimoto similarity within a set of fingerprints or between two sets of fingerprints.
 
-    Expects fingerprints to be generated by nvMolKit, with the leading dimension corresponding to
+    Expects fingerprints generated by nvMolKit by nvMolKit or as a torch tensor located on device 0, with the leading dimension corresponding to
```
evasnow1992: Duplicate of "by nvMolKit"

tests/test_similarity.cpp
```diff
@@ -162,659 +164,162 @@ cuda::std::span<const OtherT> castAsSpanOfSmallerType(const AsyncDeviceVector<T>
 }
 
 // --------------------------------
-// Tanimoto similarity tests
```
evasnow1992, line 165: It looks like the SimilarityTestFixture defined on line 136 and SimilarityParamTestFixture defined on line 143 above are no longer used after all the cleanup on this script.

evasnow1992 (approved):
Thank you for cleaning up similarity APIs! Only two minor comments to address. Otherwise look good to me.

tests/test_similarity.cpp
scal444: Done

nvmolkit/similarity.py
scal444: Fixed!

PR #13, opened 2025-09-26
Add CPU-result segmented Tanimoto/Cosine similarity

scal444:
Similarity done in chunks on the GPU, and sections are rotated back to the CPU result buffer via pinned memory. Two pinned memory buffers so that the next CUDA memcpy can overlap the pinned -> result copy.


src/similarity.cpp
```diff
@@ -43,45 +36,252 @@ constexpr int                    kBitsPerByte               = 8;
 
 AsyncDeviceVector<double> crossTanimotoSimilarityGpuResult(const cuda::std::span<const std::uint32_t> bits,
```
evasnow1992, line 37: The keyword 'const' has been removed from the function declaration in similarity.h. Should we update all related function signatures here or add 'const' back in the header file?

scal444: The consts before had been ignored. In a declaration, const is only applied to reference or pointer types.

So while there's a difference between `const int*` and `int*`, there's no difference between `int` and  `const int` or any other value type **in the header**. It does matter in the implementation.

evasnow1992 (approved):

PR #16, opened 2025-10-01
Remove 256 atom limit for minimizer

scal444:
A fallback kernel is implemented that doesn't use the cached shared memory. 

Added a hessian update benchmark script

### DIFF
--- a/tests/test_mmff.cu
+++ b/tests/test_mmff.cu
@@ -14,7 +14,6 @@
 // limitations under the License.
 
 #include <ForceField/ForceField.h>
-#include <ForceField/MMFF/BondStretch.h>
 #include <gmock/gmock.h>
 #include <GraphMol/DistGeomHelpers/Embedder.h>
 #include <GraphMol/FileParsers/FileParsers.h>
@@ -22,7 +21,6 @@
 #include <GraphMol/ForceFieldHelpers/MMFF/AtomTyper.h>
 #include <GraphMol/ForceFieldHelpers/MMFF/Builder.h>
 #include <GraphMol/ForceFieldHelpers/MMFF/MMFF.h>
-#include <GraphMol/SmilesParse/SmilesParse.h>
 #include <gtest/gtest.h>
 
 #include <filesystem>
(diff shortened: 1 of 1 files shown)


tests/test_mmff.cu
```diff
@@ -1542,25 +1543,84 @@ TEST(MMFFMultiGPU, MultiGPUSpecificIds) {
...
+  RDKit::MolOps::sanitizeMol(*mol);
+  RDKit::MolOps::addHs((*mol), /*explicitOnly=*/false, /*addCoords=*/true);
+  return mol;
+}
+
+TEST(MMFFAllowsLargeMol, LargeMoleculeInterleavedOptimizes) {
+  constexpr double bondLength     = 1.0;
+  constexpr double hydrogenOffset = 1.0;
+
+  auto small2 = createHydroCarbon(3, bondLength, hydrogenOffset);
```
evasnow1992: Is creating small2 for testing createHydroCarbon? Otherwise it's not used anywhere later.

tests/test_mmff.cu
```diff
@@ -1542,25 +1543,84 @@ TEST(MMFFMultiGPU, MultiGPUSpecificIds) {
...
+  ASSERT_NE(small2, nullptr);
   ASSERT_NE(big, nullptr);
-  ASSERT_GT(big->getNumAtoms(), 256u);
 
-  // Give each a conformer for MMFF
-  small1->addConformer(new RDKit::Conformer(small1->getNumAtoms()));
-  small2->addConformer(new RDKit::Conformer(small2->getNumAtoms()));
-  big->addConformer(new RDKit::Conformer(big->getNumAtoms()));
+  std::vector<std::unique_ptr<RDKit::RWMol>> rdkitRefs;
+  // rdkitRefs.push_back(std::make_unique<RDKit::RWMol>(*big));
```
evasnow1992: redundant comment?

tests/test_mmff.cu
scal444: Thanks, for got to add back in the small molecules, the test case I wanted was {small, large, small}. I also added a random kick to the coordinates, because we were starting from fully symmetric which causes issues reproducing between rdkit/nvmolkit because it's a saddle point.

tests/test_mmff.cu
```diff
@@ -1542,25 +1541,86 @@ TEST(MMFFMultiGPU, MultiGPUSpecificIds) {
...
 
-TEST(MMFFAtomLimit, OversizedMoleculeInterleavedThrows) {
-  // Small molecules
-  auto small1 = std::unique_ptr<RDKit::RWMol>(RDKit::SmilesToMol("CCCCCC"));
-  auto small2 = std::unique_ptr<RDKit::RWMol>(RDKit::SmilesToMol("CCC"));
-  ASSERT_NE(small1, nullptr);
-  ASSERT_NE(small2, nullptr);
+std::unique_ptr<RDKit::RWMol> createHydroCarbon(const int    numCarbons,
+                                                const double bondLength,
+                                                const double hydrogenOffset) {
```
evasnow1992, line 1546: @scal444 just noticed that hydrogenOffset is not used inside this function.

scal444: Ack will remove next time we touch this code

PR #17, opened 2025-10-02
Enable full ETKDG variant support

evasnow1992:
This PR implements comprehensive support for all ETKDG variants in nvMolKit, aligning the behavior with RDKit's reference implementation.


src/etkdg.cpp
```diff
@@ -248,26 +248,43 @@ void embedMolecules(const std::vector<RDKit::ROMol*>&           mols,
       stages.push_back(
         std::make_unique<detail::FirstMinimizeStage>(constMolPtrs, batchEargs, paramsCopy, context, streamPtr));
       stages.push_back(std::make_unique<detail::ETKDGTetrahedralCheckStage>(context, batchEargs, dim, streamPtr));
-      auto chiralStage =
-        std::make_unique<detail::ETKDGFirstChiralCenterCheckStage>(context, batchEargs, dim, streamPtr);
-      auto& chiralStageRef = *chiralStage;
-      stages.push_back(std::move(chiralStage));
+
+      // Only add first chiral check if enforceChirality is enabled
+      detail::ETKDGFirstChiralCenterCheckStage* chiralStagePtr = nullptr;
```
scal444, line 253: Noting for followup (probably before we release v0.2). We now have variable length pipelines and errors, so if we want to report failure rates to the user we also need to pass a description of each stage (probably as an enum).

src/etkdg.cpp
```diff
@@ -248,26 +248,43 @@ void embedMolecules(const std::vector<RDKit::ROMol*>&           mols,
...
-      // This is a pass-through, don't need to set the stream
-      stages.push_back(std::make_unique<detail::ETKDGFinalChiralCenterCheckStage>(chiralStageRef));
-      stages.push_back(std::make_unique<detail::ETKDGChiralDistMatrixCheckStage>(context, batchEargs, dim, streamPtr));
-      stages.push_back(
-        std::make_unique<detail::ETKDGChiralCenterVolumeCheckStage>(context, batchEargs, dim, streamPtr));
-      stages.push_back(std::make_unique<detail::ETKDGDoubleBondStereoCheckStage>(context, batchEargs, dim, streamPtr));
+
+      if (paramsCopy.enforceChirality) {
+        // Only add final chiral check if we have a first chiral stage to reference
+        if (chiralStagePtr != nullptr) {
```
scal444: When would this ever be nullptr with enforceChirality set?

tests/test_etkdg_etk_minimize.cu
```diff
@@ -82,13 +82,15 @@ void perturbConformer(RDKit::Conformer& conf, const float delta = 0.1, const int
...
-                                       std::vector<double>*                    posOut       = nullptr) {
+                                       bool                                    minimizerTol      = 1e-3,
+                                       double*                                 posOverride       = nullptr,
+                                       std::vector<double>*                    posOut            = nullptr,
+                                       bool                                    useBasicKnowledge = true) {
   std::vector<double> results;
   results.reserve(mols.size());
   int startIdx = 0;
   for (size_t i = 0; i < mols.size(); ++i) {
+    std::cout << "molecule index: " << i << std::endl;
```
scal444: Leftover debugging?

tests/test_distgeom_kernels.cu
```diff
@@ -1274,7 +1241,8 @@ class ETK3DGpuTestFixture : public ::testing::Test {
   std::vector<int>                                    atomStartsHost;
 };
 
-void ETK3DGpuTestFixture::setUpSystems(const std::vector<RDKit::ROMol*>& molBatch) {
+void ETK3DGpuTestFixture::setUpSystems(const std::vector<RDKit::ROMol*>&    molBatch,
```
scal444, line 1244: Should this be named ETDG rather than ETK?

nvmolkit/tests/test_embed_molecules.py
```diff
@@ -121,81 +131,97 @@ def compare_conformers_rmsd(rdkit_mols, nvmolkit_mols, rmsd_threshold=0.2, min_m
 
         rdkit_conf_count = rdkit_mol.GetNumConformers()
         nvmolkit_conf_count = nvmolkit_mol.GetNumConformers()
-        
+
         if rdkit_conf_count == 0 or nvmolkit_conf_count == 0:
```
scal444, line 135: I don't have a problem with reformatting the python files, but I'm a bit worried why when I run the formatter it also doesn't pick up on these changes. Can you let me know exactly how you're invoking ruff and which version it is?

scal444 (commented):
This looks awesome! Really clean work.

scal444:
One more comment, it's not one of the official presets, but I think there's also just a DG where both ET and K terms are turned off.

src/etkdg.cpp
evasnow1992: This extra checking is redundant given it's inside the enforceChirality block. Removing it.

tests/test_etkdg_etk_minimize.cu
evasnow1992: Yes, removing it

tests/test_distgeom_kernels.cu
evasnow1992: If you are referring to the name of this test fixture, it is used for testing all the ET and K terms that are handled by construct3Dcontrib and used by the ETK_minimize stage.
The ETDG parameter is used in some tests to trigger the plain 3D constructor, while the other tests use the default ETKDGv3 parameter set.

nvmolkit/tests/test_embed_molecules.py
evasnow1992: Most of the format issues I corrected this time were originally introduced when I added the Python APIs for ETKDG and MMFF. I didn't have a formatter locally, and our CI pipeline didn't catch them when I merged those changes last time.
I've since added a Ruff extension on my VS Code, which automatically highlights these format issues in my scripts. This is how I identified them this time. It would be a great improvement if we could also include a lint format checker in our CI pipeline to prevent these issues in the future.

scal444: Ok sure, good enough for now.

src/etkdg.cpp
evasnow1992: Based on my examination of our current implementation, enabling pipeline variant stage names will require updates to etkdg_impl, etkdg, etkdg_kernel, and related test files. Additionally, failure information is not yet exposed to the Python API (it currently uses a default nullptr) but is only available via the C++ API. Since these changes will be non-trivial, I will create an additional PR for them instead of updating the current PR.

scal444 (approved):

PR #18, opened 2025-10-06
Add exception handling in all library OpenMP loops

### LINKED ISSUE #9: Exceptions are not propagating to python properly
Exceptions are not bubbling up properly to python when in OpenMP loops such as Morgan Fingerprinting. We need to handle exceptions in those loops by saving then throwing once out of the loop.

scal444:
Added a new class which just stores the first seen exception then throws at the end of the loop.

Additionally, relaxed an energy check in test_etkdg that was failing. We have a few very simple molecules which can reach zero energy with some ETKDG variants.

Ran formatter which changed test_distgeom_kernels slightly.

Fixes #9

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -29,7 +29,6 @@ Checks: '
    -modernize-use-trailing-return-type,
   mpi-*,
   openmp-*,
-    -openmp-exception-escape,
     -openmp-use-default-none
   performance-*,
   readability-*,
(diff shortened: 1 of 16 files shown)


scal444:
@evasnow1992 sorry for the big diff but most of it is just indentation.

evasnow1992 (approved):
Thank you for adding these changes to propagate errors from multithreading. The changes look good to me.

PR #19, opened 2025-10-07
Prepare for v0.2 and conda release

scal444:
* Bumped versions in pyproject and python code
* Fixed warnings that showed up in some conda builds. Apparently constructors with `= default` don't need __host__ __device__ annotations in modern CUDA compilers, they'll generate both automatically.
* Added some helper scripts to verify conda builds.


evasnow1992 (approved):
Thank you for boosting the package version and adding scripts for verifying conda builds. Changes look good to me.

PR #20, opened 2025-10-08
Fix bugs with memory cleanup and large similarity arrays

scal444:
* Added destructor to py array so that cudaFree is properly called. Added a test for it
* Changed 32 bit ints to 64 bit ints in pybind to avoid potential overflow


scal444:
build

evasnow1992 (approved):
Thank you for adding the array destructor. Changes look good to me.

PR #27, opened 2025-10-20
Update C++ versions and bump nvMolKit version for v0.3 dev

scal444:
Also bumped cmake required version.

Note that the v0.2 branch had the version updated separately so isn't missing this, at least for the python side.


evasnow1992 (approved):
Thank you for bumping the version. Changes look good for me!

PR #35, opened 2025-10-27
Remove old threading count check

### LINKED ISSUE #31: On low cpu machine, batchesPerGpu (4) cannot exceed numThreads (2)
In testing built wheels on CI, some tests fail due to
```
FAILED nvMolKit/nvmolkit/tests/test_embed_molecules.py::test_embed_molecules_allows_large_molecule_interleaved - ValueError: batchesPerGpu (4) cannot exceed numThreads (2)
FAILED nvMolKit/nvmolkit/tests/test_embed_molecules.py::test_embed_molecules_prune_rmsthresh - ValueError: batchesPerGpu (4) cannot exceed numThreads (2)
FAILED nvMolKit/nvmolkit/tests/test_mmff_optimization.py::test_error_case_throws_properly - ValueError: batchesPerGpu (4) cannot exceed numThreads (2)
```

```
    def test_embed_molecules_prune_rmsthresh():
        mols = [Chem.MolFromSmiles("c1ccccc1"), Chem.MolFromSmiles("C" * 30)]
        params = EmbedParameters()
        params.useRandomCoords = True
        params.pruneRmsThresh = 0.5
>       embed.EmbedMolecules(mols, params, confsPerMolecule=5)
[...]
        # Call the C++ implementation
>       _embedMolecules.EmbedMolecules(molecules, params, confsPerMolecule, maxIterations, native_options)
E       ValueError: batchesPerGpu (4) cannot exceed numThreads (2)
```

This is likely due to the fact that the default embed hardware options have `batchesPerGpu = -1`, and `-1` falls back to 4:
https://github.com/NVIDIA-Digital-Bio/nvMolKit/blob/main/src/etkdg.cpp#L79
and then raises error due to the check right below.

I was surprised that none of the bfgs tests failed: https://github.com/NVIDIA-Digital-Bio/nvMolKit/blob/main/src/minimizer/bfgs_mmff.cpp#L52-L54

Probably we can have the same type of etkdg logic as that in bfgs.

scal444:
The CPU vs GPU driving threads are decoupled and we shouldn't check for consistency here. In the extreme case of a 1 CPU system, it's valid to have 1 CPU thread for preprocessing and multiple for GPU dispatch, if not ideal. 

Fixes #31

### DIFF
--- a/src/etkdg.cpp
+++ b/src/etkdg.cpp
@@ -81,11 +81,6 @@ void embedMolecules(const std::vector<RDKit::ROMol*>&           mols,
     hardwareOptions.preprocessingThreads == -1 ? omp_get_max_threads() : hardwareOptions.preprocessingThreads;
   const int batchSize = hardwareOptions.batchSize == -1 ? 500 : hardwareOptions.batchSize;
 
-  // Validate batchesPerGpu
-  if (batchesPerGpu > numThreads) {
-    throw std::invalid_argument("batchesPerGpu (" + std::to_string(batchesPerGpu) + ") cannot exceed numThreads (" +
-                                std::to_string(numThreads) + ")");
-  }
   if (batchesPerGpu <= 0) {
     throw std::invalid_argument("batchesPerGpu must be greater than 0");
   }


evasnow1992 (approved):
Thank you for addressing this stale check.

PR #38, opened 2025-10-30
Use fast math and fix compiler flag appending issue

scal444:
The warning suppressions had been appended each invocation, which was harmless but led to a bunch of repetition in nvcc invocations.

For fast math, this doesn't break our tests because we're currently mostly in double precision, but have future single precision work that will be affected by it.


evasnow1992 (approved):
Thank you for cleaning flag duplications and refining the fast math scope.

PR #39, opened 2025-10-31
Implement Butina Clustering

scal444:
This PR also contains a small fix for size zero device vectors with explicit constructor, and allows local pip to build with non-release version if specified.


src/butina.cu
```diff
@@ -0,0 +1,325 @@
...
+    }
+  }
+}
+
+//! Kernel to bump the cluster index if the last cluster assigned was large enough. The edge case is for if we hit the <
+//! 3 criteria.
+__global__ void bumpClusterIdxKernel(int* clusterIdx, const int* lastClusterSize) {
+  if (const auto tid = static_cast<int>(threadIdx.x); tid == 0) {
+    if (*lastClusterSize >= kMinLoopSizeForAssignment) {
+      clusterIdx[tid] += 1;
```
evasnow1992: Minor point. For better readability and better consistency with the how clusterIdx is accessed in the butinaWriteClusterValue kernel above (through *clusterIdx), may consider updating this to (*clusterIdx) += 1; or add a comment like "clusterIdx has size 1"

src/butina.cu
```diff
@@ -0,0 +1,325 @@
...
+namespace {
+
+//! CUB Op to threshold a distance matrix into a hit matrix
+struct ThresholdOp {
+  const double* matrix;
+  uint8_t*      hits;
+  double        cutoff;
+  ThresholdOp(const double* m, uint8_t* h, double c) : matrix(m), hits(h), cutoff(c) {}
+
+  __device__ void operator()(const std::size_t idx) const { hits[idx] = (matrix[idx] < cutoff); }
```
evasnow1992: In the rdkit's current python implementation, neighbors are identified with `neighbor_lists = [np.where(dist_matrix[i] <= distThresh)[0].tolist() for i in range(nPts)]`, i.e. with <= instead of <. Just wondering if we want to replicate rdkit's behavior. Algorithm-wise the current implementation is valid as butina clustering.

tests/test_butina.cpp
```diff
@@ -0,0 +1,198 @@
...
+      distances[(col * nPts) + row] = value;
+    }
+  }
+  return distances;
+}
+
+std::vector<uint8_t> makeAdjacency(const std::vector<double>& distances, int nPts, double cutoff) {
+  std::vector<uint8_t> adjacency(distances.size(), 0);
+  for (size_t idx = 0; idx < distances.size(); ++idx) {
+    adjacency[idx] = distances[idx] < cutoff ? 1U : 0U;
```
evasnow1992: Need update here if we decide to switch from < to <= in butina.cu

nvmolkit/clustering.py
```diff
@@ -0,0 +1,49 @@
...
+
+""" Contains GPU-accelerated Butina clustering implementation. """
+
+import torch
+
+from nvmolkit import _clustering
+from nvmolkit._arrayHelpers import *  # noqa: F403
+from nvmolkit.types import AsyncGpuResult
+
+def butina(distance_matrix: AsyncGpuResult | torch.Tensor, cutoff: float) -> AsyncGpuResult:
```
evasnow1992, line 24: Not for this MR, but we may also add a wrapper function takes and returns CPU inputs.

evasnow1992 (approved):
Thank you for adding the GPU-accelerated butina clustering. The implementation is solid and valid. Only minor comments. No further changes are mandatory for this MR.

nvmolkit/clustering.py
scal444: In this case I'd say calling result.numpy() would be sufficient. We have CPU result APIs when it's a different path, such as when we're doing optimizations for GPU memory limitations

src/butina.cu
scal444: Done

tests/test_butina.cpp
scal444: Done, and in python tests

src/butina.cu
scal444: Done

PR #40, opened 2025-10-31
Compatibility layer for different CUB versions

scal444:
Also removed some typenames that are not needed with modern C++/CUDA

### DIFF
--- a/src/utils/CMakeLists.txt
+++ b/src/utils/CMakeLists.txt
@@ -50,5 +50,5 @@ target_link_libraries(openmp_helpers PRIVATE OpenMP::OpenMP_CXX)
 target_include_directories(openmp_helpers INTERFACE ${CMAKE_CURRENT_SOURCE_DIR})
 
 add_library(cub_helpers INTERFACE)
-target_link_libraries(host_vector INTERFACE CUDA::cudart)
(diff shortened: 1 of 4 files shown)


src/utils/CMakeLists.txt
```diff
@@ -48,3 +48,7 @@ target_include_directories(nvtx INTERFACE ${CMAKE_CURRENT_SOURCE_DIR})
 add_library(openmp_helpers openmp_helpers.cpp)
 target_link_libraries(openmp_helpers PRIVATE OpenMP::OpenMP_CXX)
 target_include_directories(openmp_helpers INTERFACE ${CMAKE_CURRENT_SOURCE_DIR})
+
+add_library(cub_helpers INTERFACE)
```
evasnow1992, line 52: Do we need to link cub_helpers to bfgs library defined in src/minimizer/CMakeLists.txt in order to use it in bfgs_minimize.cu?

evasnow1992 (approved):
Thank you for adding the function alias. Just one comment regarding linking the new library to bfgs library. The other changes look good to me.

src/utils/CMakeLists.txt
scal444: Done

Issue #3, opened 2025-09-11
Build fails with CUDA 12.8 + libcudacxx: cuda/std/span not found + massive cub/thrust conflicts


cab-tokoron:
**Build fails with CUDA 12.8 + libcudacxx: cuda/std/span not found + massive cub/thrust conflicts**

I'm trying to build nvMolKit in a clean conda environment using CUDA 12.8 and libcudacxx installed to `/usr/local/cuda/include`. Despite confirming that `cuda/std/span` exists, the build fails with hundreds of errors related to `cub`, `thrust`, and libcudacxx internal types.

### Environment:
- OS: Linux Mint 21.3
- GPU: NVIDIA RTX 2060
- CUDA Toolkit: 12.8
- libcudacxx: cloned from GitHub and installed to `/usr/local/cuda/include`
- Python: 3.10 (via conda)
- nvMolKit: latest from GitHub
- CMake: 3.26
- Compiler: nvcc from CUDA 12.8

### What I tried:
- Verified `cuda/std/span` exists
- Added `include_directories(/usr/local/cuda/include/cuda/std)` to CMakeLists.txt
- Tried setting `CMAKE_CUDA_FLAGS` and `add_definitions(-DCUB_DISABLE)` etc.
- Commented out `#include <cuda/std/span>` in `device_vector.h`
- Attempted to isolate libcudacxx headers inside the conda environment
- All attempts resulted in hundreds of errors like:

/usr/local/cuda/include/cub/thread/thread_reduce.cuh(71): error: namespace "cuda::std" has no member "__accumulator_t" 
/usr/local/cuda/include/cuda/std/detail/libcxx/include/cmath(679): error: duplicate specifier in declaration 
/usr/local/cuda/include/thrust/tuple.h(216): error: expected a ";" 
...

### Summary:
It seems nvMolKit currently cannot be built against CUDA 12.8 + libcudacxx due to deep conflicts between libcudacxx and other CUDA headers (cub, thrust, etc.). If there's a recommended configuration or workaround, I'd be happy to test it.

Thanks for your work on this project!

cab-tokoron:
Thanks for the follow-up and for trying to reproduce the issue.

To clarify my setup:

- OS: Linux Mint 21.3 (based on Ubuntu 22.04)
- CUDA Toolkit: 12.8 (installed via `.run` installer from NVIDIA's website)
[...]

PR #43, opened 2025-11-05
Add automatic detection and support for external CCCL

evasnow1992:
Add automatic CMake detection for external CCCL installations (e.g., from conda), enabling users to leverage newer Thrust/CUB features while maintaining backward compatibility.

- Create `cccl_interface` library that wraps CCCL::CCCL when version >= 3.0.0
- Add `NVMOLKIT_HAS_CCCL_GE_3` macro based on CMake detection (not CUDART_VERSION)
- Use `cuda::maximum<>`/`cuda::std::plus<>` for CCCL >= 3.0.0, `cub::Max`/`cub::Sum` otherwise
- Suppress deprecation warnings for CCCL 2.x compatibility
- Simplify linking and add clear status messages about CCCL source

Tested with: no CCCL (CUDA 12.6 bundled), conda CCCL 3.1.1, 3.0.0, and 2.8.5.


scal444:
Looks great! @pechersky FYI, finishing #3

PR #47, opened 2025-11-24
CCCL cleanup and minor utilities improvement

scal444:
* Ran cmake-format, most of the changes here are formatting
* Added hint for CCCL to look in the CUDA installation
* Added extra conditional on CCCL macros
* Changed host_vector to only resize on increase, not decrease

### DIFF
--- a/src/utils/cub_helpers.cuh
+++ b/src/utils/cub_helpers.cuh
@@ -18,9 +18,10 @@
 
 #include <cub/cub.cuh>
 
-#ifdef NVMOLKIT_HAS_CCCL_GE_3
+#if defined(NVMOLKIT_HAS_CCCL_GE_3) || CUDART_VERSION >= 12090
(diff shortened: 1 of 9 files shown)


src/utils/cub_helpers.cuh
```diff
@@ -18,17 +18,19 @@
 
 #include <cub/cub.cuh>
 
-#ifdef NVMOLKIT_HAS_CCCL_GE_3
+#if defined(NVMOLKIT_HAS_CCCL_GE_3) || CUDART_VERSION >= 12090
 // CCCL >= 3.0.0 provides modern C++ functional operators
```
evasnow1992, line 22: Thank you for adding this. I also realized we should expand this condition to include checking cuda toolkit version when I was working on cub argMax

evasnow1992 (approved):
Thank you for cleaning up the format, updating the CCCL version checking condition, and fixing the resize function.

PR #48, opened 2025-11-24
Move most MMFF Kernel logic to mmff_kernels_device.cuh and reduce precision where possible. 

scal444:
Templated some helper functions to support float, changed a few data types in storage, and relaxed precision requirements on tests.


src/forcefields/mmff_kernels_device.cuh
```diff
@@ -0,0 +1,667 @@
...
+                                                   const double  R_ij_star,
+                                                   const double  wellDepth) {
+  // Note, this kernel is quite sensitive, any downcasting to fp32 causes significant drift.
+  double R_ij_star2 = R_ij_star * R_ij_star;
+  double R_ij_star7 = R_ij_star2 * R_ij_star2 * R_ij_star2 * R_ij_star;
+
+  const double epsilon = wellDepth;
+
+  const double distSquared = distanceSquared(pos, idx1, idx2);
+  const double dist        = sqrtf(distSquared);
```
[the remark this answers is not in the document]
scal444: This is deliberate, casts to float, use float sqrt (much faster), cast back to double because double is needed before and after.

tests/test_mmff.cu
```diff
@@ -652,12 +667,12 @@ TEST_F(MMffGpuEdgeCases3Atoms, ZeroThetaAngleStretchBend) {
   ASSERT_EQ(referenceForceField_->contribs().size(), 1);
   double wantEnergy = referenceForceField_->calcEnergy(positions.data());
   double gotEnergy  = getEnergyTerm(systemDevice, FFTerm::StretchBend);
-  EXPECT_NEAR(gotEnergy, wantEnergy, 1e-3);
+  EXPECT_NEAR(gotEnergy, wantEnergy, 1e-1);
```
evasnow1992: Not required, but since we are cleaning up these tolerance thresholds, do we also want to define some const variables for these 1e-1 energy and grad thresholds?

evasnow1992 (approved):
Thank you for consolidating these helper functions and enabling using floats. Only one minor comment for test tolerance variables.

tests/test_mmff.cu
scal444: Done

PR #50, opened 2025-11-25
Improve ETKDG Setup times via caching and data reuse

scal444:
* A single minimizer is instantiated per thread and reused
* First and 4th dim minimization now reuse data. 4th dim is a wrapper around 1st dim. Changed kernels that use the different weights to take in double parameters, consolidating what had been identical arrays.
* Added a cache of parameters to pass in to all minimization stages, so that molecules aren't processed multiple times
* Preallocate parameters based on the first molecule so we don't constantly resize


benchmarks/etkdg_ff_bench.cpp
```diff
@@ -108,14 +108,14 @@ void benchNvMolKitBatch(const std::vector<RDKit::ROMol*>& mols, const int size)
   ankerl::nanobench::Bench().run("nvMolKit ETKDG calc energy, nmols = " + std::to_string(size), [&] {
     systemDevice.energyBuffer.zero();
     systemDevice.energyOuts.zero();
-    ankerl::nanobench::doNotOptimizeAway(computeEnergy(systemDevice, atomStartsDevice, positionsDevice));
+    ankerl::nanobench::doNotOptimizeAway(computeEnergy(systemDevice, atomStartsDevice, positionsDevice, 1.0, 1.0));
```
evasnow1992: Should the added values be 1.0 and 0.1?

evasnow1992 (approved):
Thank you for consolidating the minimization stages and adding cache. Only one clarification question.

benchmarks/etkdg_ff_bench.cpp
scal444: For benching the energy/grad it won't make a difference but changed to be consistent.

PR #51, opened 2025-11-25
Scratch buffers for common copies

scal444:
* Add code to reuse an ETKDG context to not reallocate
* Add per thread scratch buffers for copying statuses and positions, and failure checks


evasnow1992 (approved):
Thank you for enabling reusing ETKDG drivers and adding pinned buffers

PR #53, opened 2025-12-01
Add e2e BFGS minimization in single kernel, apply to MMFF

scal444:
MMFF can now be run end to end, with 1 block handling one molecule. 

 * Used recently refactored device energy/gradient kernels
 * Migrated all BFGS steps to single block
 * Made small molecule optimization to use shared memory for common buffers.
 * Old backend still works


src/minimizer/bfgs_minimize.cu
```diff
@@ -18,8 +18,12 @@
 
 #include "bfgs_hessian.h"
 #include "bfgs_minimize.h"
+#include "bfgs_minimize_permol_kernels.h"
```
evasnow1992, line 21: Will this create a circular dependency? As `bfgs_minimize_permol_kernels.h` also imports `bfgs_minimize.h`. It may not result in a compile error with all the `#ifndef...#define...` pairs, but may make it harder for later maintenance.

src/minimizer/bfgs_minimize_permol_kernels.h
```diff
@@ -0,0 +1,47 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef NVMOLKIT_BFGS_MINIMIZE_PERMOL_KERNELS_H
+#define NVMOLKIT_BFGS_MINIMIZE_PERMOL_KERNELS_H
+
+#include <cuda_runtime.h>
+
+#include "bfgs_minimize.h"
```
evasnow1992: The potential circular dependency issue mentioned above

src/minimizer/bfgs_mmff.cpp
```diff
@@ -143,35 +194,65 @@ std::vector<std::vector<double>> MMFFOptimizeMoleculesConfsBfgs(std::vector<RDKi
...
+                               systemHost.indices.atomStarts,
+                               systemDevice.indices.atomStarts,
+                               systemDevice.positions,
+                               systemDevice.grad,
+                               systemDevice.energyOuts,
+                               systemDevice.energyBuffer,
+                               eFunc,
+                               gFunc);
+      } else {
+        auto terms         = nvMolKit::MMFF::toEnergyForceContribsDevicePtr(systemDevice);
```
evasnow1992, line 229: I saw these two functions are defined in `mmff.h` but did not find where `mmff.h` is included. Looks like it's not included in `bfgs_mmff.h` or `bfgs_minimize.h` to give `bfgs_mmff` access? I may have missed it somewhere.

evasnow1992 (approved):
A few clarification questions on header file dependencies. Changes look good to me. Thank you for adding molecule level MMFF e2e minimization!

src/minimizer/bfgs_minimize.cu
scal444: Thanks, fixed by moving it out into bfgs_types.h

src/minimizer/bfgs_mmff.cpp
scal444: Looks like it was being included via the flattened build import. Changed to import directly.

PR #54, opened 2025-12-02
Add a block-per-molecule path for ETKDG

scal444:


evasnow1992 (approved):
Thank you for adding per-molecule path for ETKDG. Changes look good to me.

PR #55, opened 2025-12-03
Fix stream ordering bug in device pointers

scal444:
These were being allocated on the default stream, meaning that very infrequently it was possible to access them before allocation completed. This may have been another source of the user reported sync errors.


evasnow1992 (approved):
Thank you for identifying and fixing this issue.

PR #56, opened 2025-12-05
Fix various possible memory and overflow issues

scal444:
The main problem fixed here is the use of int16 indices was overflowing on larger batches. Some of these are safe to downcast because they use internal indexing to a molecule, but some are globally indexed. Removed all of them to avoid sharp edges.

Fixed several less likely issues where a memcpy H2D could have the host go out of scope and get destructed before the memcpy was finished. 

Fixed a possible invalid API call if DeviceVector zero() was called on an empty array


evasnow1992 (approved):
Thank you for identifying and fixing these issues

PR #57, opened 2025-12-05
Add small cluster neighborlist path to Butina implementation

scal444:
We switch once the max cluster size is 8 (can revisit that later) and try to find isolated clusters which will never be part of a larger one, reducing iteration counts. 

Also one tiny cmake fix


evasnow1992 (approved):
Thank you for making this enhancement!

PR #59, opened 2025-12-05
Remove separate dispatch for per molecule kernel sizes. Now choose one kernel based on max size.

scal444:
Sending off kernel32 then kernel64 was a pessimization. We lose a tiny bit of per-kernel perf from this but we need batches to send off at once, not sequentially, and we can't afford the communication overhead of instantiating more concurrent streams.

Also fixed a leftover debugging statement that was causing the MMFF benchmark to run slow by calculating RDKit energies when not necessary. Did not affect result reporting, just did extraneous work in between benchmarks.

### DIFF
--- a/src/minimizer/bfgs_minimize.cu
+++ b/src/minimizer/bfgs_minimize.cu
@@ -517,25 +517,23 @@ void BfgsBatchMinimizer::initialize(const std::vector<int>& atomStartsHost,
 
   activeSystemIndices_.resize(numSystems_);
   allSystemIndices_.resize(numSystems_);
-  std::vector<int> activeSystemIndicesHost(numSystems_);
-  std::iota(activeSystemIndicesHost.begin(), activeSystemIndicesHost.end(), 0);
-  allSystemIndices_.setFromVector(activeSystemIndicesHost);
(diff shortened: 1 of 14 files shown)


src/minimizer/bfgs_minimize.cu
```diff
@@ -406,9 +404,7 @@ BfgsBatchMinimizer::BfgsBatchMinimizer(const int    dataDim,
     inverseHessian_.setStream(stream_);
     hessDGrad_.setStream(stream_);
     scratchBuffersDevice_.setStream(stream_);
-    for (auto& bin : perMolBinListsDevice_) {
-      bin.setStream(stream_);
-    }
+    activeMolIdsDevice_.setStream(stream_);
```
evasnow1992, line 402: Is this setStream redundant giving that it's already set above on line 376 in `if (backend_ == BfgsBackend::PER_MOLECULE)` block?

evasnow1992 (approved):
LGTM. Only one clarification question.

src/minimizer/bfgs_minimize.cu
scal444: Fixed, thanks. Kept the second one for simplicity.

PR #61, opened 2025-12-05
Use shared memory positions in per molecule small kernels

scal444:
We'd migrated gradients and other shared buffers but were still using global arrays for positions.

### DIFF
--- a/src/minimizer/bfgs_minimize_permol_kernels.cu
+++ b/src/minimizer/bfgs_minimize_permol_kernels.cu
@@ -620,16 +620,16 @@ __global__ void bfgsMinimizeKernel(const int               numIters,
       }
       __syncthreads();
 
-      // Compute energy at perturbed position
+      // Compute energy at perturbed position (use scratchPos which has the perturbed coordinates)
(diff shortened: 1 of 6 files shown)


src/minimizer/bfgs_minimize_permol_kernels.cu
```diff
@@ -620,16 +620,16 @@ __global__ void bfgsMinimizeKernel(const int               numIters,
       }
```
stslxg-nv: We can also remove the copying to `globalPos` in line search before energy function and after line search.

stslxg-nv: But maybe we need to copy `localPos` back to `globalPos` at the end if the result of `positions` is used?

scal444: Good point. That had been happening below outside of the line search kernel, so removing the extra copy is fine. But the extra copy outside was extraneous. Added another commit, probably easiest to see from that diff

stslxg-nv (approved):
LGTM

PR #62, opened 2025-12-08
Dynamic neighborlist for Butina clustering

scal444:
* In the inner loop, the neighborlist is set once and then dynamically reduced
* Made the size of the neighborlist variable. Helps performance a lot to get up to ~64, since we switch to the neighborlist earlier
* Removed the cub bulk for. It was doing weird things in nsys profiles
* Removed doublet kernel since the neighborlist loop does doublets. Removed singlet and doublet sentinel values
* Parallelized singlet assignment


evasnow1992 (approved):
LGTM. Thank you for enabling dynamic neighborlist with changeable maximum neighbor list size.

PR #64, opened 2025-12-08
Add hybrid BFGS backend that will choose at runtime between the backends based on size. 

scal444:
Also includes some clang-format


evasnow1992 (approved):
Thank you for consolidating different backends and enabling automatic selection. LGTM.

PR #66, opened 2025-12-08
Final Butina PR for v0.3 - remove strict/non strict path, sort keys ourselves

scal444:
This PR removes the strict/non strict divergence, using relaxed only and then re-sorting cluster IDs at the end to enforce strictness. 

* Improved speed of benchmark verification check.
* Cast to size t a few places where we could have overflow
* Added back in the bump counts kernel. The "optimization" merging it with the assignment kernel was a data race


evasnow1992 (approved):
Thank you for cleaning the strict/non strict paths. LGTM.

PR #67, opened 2025-12-08
Use CUB DeviceReduce::ArgMax for Butina clustering with CCCL version detection

evasnow1992:
Replace custom ArgMax kernel in Butina clustering with CUB's DeviceReduce::ArgMax when CCCL >= 2.8.0, while maintaining backward compatibility for older CCCL versions.

Also clean up some unused variables

### DIFF
--- a/src/butina.cu
+++ b/src/butina.cu
@@ -290,6 +290,7 @@ __global__ void assignSingletonIdsKernel(const cuda::std::span<int> clusters, co
   }
 }
 
+#if CUB_VERSION < 200800
 constexpr int argMaxBlockSize = 512;
 
 //! Custom ArgMax kernel that returns the largest value and index.
@@ -322,12 +323,13 @@ __global__ void lastArgMaxKernel(const int* values, int numItems, int* outVal, i
     }
   }
 }
+#endif  // CUB_VERSION < 200800
 
 //! Helper class to run ArgMax on device data.
 //! Uses CUB's DeviceReduce::ArgMax when available (CCCL >= 2.8.0), otherwise falls back to custom kernel.
 class ArgMaxRunner {
  public:
-  ArgMaxRunner(size_t num_items, cudaStream_t stream)
+  ArgMaxRunner([[maybe_unused]] size_t num_items, cudaStream_t stream)
       : stream_(stream)
 #if CUB_VERSION >= 200800
         ,


src/minimizer/bfgs_minimize_permol_kernels.cu
```diff
@@ -404,22 +404,22 @@ template <> struct DataDimTraits<ForceFieldType::DG> {
 }  // namespace
 
 template <int MaxAtoms, bool UseSharedMem, ForceFieldType FFType, typename TermsType, typename IndicesType>
-__launch_bounds__(BLOCK_SIZE, 12) __global__ void bfgsMinimizeKernel(const int               numIters,
```
evasnow1992, line 407: This value seems to be too high for some GPU architectures (A6000 on my test instance), which caused `ptxas warning : Value of threads per SM for entry`

src/utils/cub_helpers.cuh
```diff
@@ -18,7 +18,27 @@
 
 #include <cub/cub.cuh>
 
-#if defined(NVMOLKIT_HAS_CCCL_GE_3) || CUDART_VERSION >= 12090
+#include "nvtx.h"
+
```
scal444: Since this is currently only used in Butina and requires a full kernel for the alternate path, please put this logic in the butina code to not pollute the header.

src/utils/cub_helpers.cuh
```diff
@@ -18,7 +18,27 @@
...
+// Default to 0 if not defined (no CCCL or unknown version)
+// Version encoding: MAJOR * 10000 + MINOR * 100 + PATCH
+// Examples: 2.8.0 = 20800, 3.0.0 = 30000
+#ifndef NVMOLKIT_CCCL_VERSION
+#define NVMOLKIT_CCCL_VERSION 0
```
scal444: Should we just always ensure it's defined in Cmake?

src/utils/cub_helpers.cuh
```diff
@@ -18,7 +18,27 @@
...
+#if NVMOLKIT_CCCL_VERSION >= 20800 || (NVMOLKIT_CCCL_VERSION == 0 && CUDART_VERSION >= 12090)
+#define NVMOLKIT_HAS_NEW_ARGMAX_API 1
+#else
+#define NVMOLKIT_HAS_NEW_ARGMAX_API 0
+#endif
+
+// Check for modern C++ operators support:
+// - CCCL >= 3.0.0 (detected via CMake as numeric version >= 30000), OR
+// - CUDA >= 13.0 when CCCL version detection failed (bundled CCCL >= 3.0.0)
+#if NVMOLKIT_CCCL_VERSION >= 30000 || (NVMOLKIT_CCCL_VERSION == 0 && CUDART_VERSION >= 13000)
```
scal444: 12.9 should support these operators too I think

CMakeLists.txt
```diff
@@ -21,33 +21,73 @@ set(CMAKE_CXX_STANDARD 20)
...
+  message(STATUS "CCCL not found, using bundled CUDA headers")
 endif()
 
 set(CMAKE_MODULE_PATH "${CMAKE_MODULE_PATH};${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 
 set(CMAKE_BUILD_TYPE
     ${CMAKE_BUILD_TYPE}
-    CACHE STRING "Choose type of build, options are: Debug RelWithDebInfo Release asan tsan ubsan" FORCE)
+    CACHE
+      STRING
```
scal444, line 48: Is this cmake formatted correctly? Surprised it hasn't been corrected by now, and it doesn't format when I run it.

CMakeLists.txt
```diff
@@ -21,33 +21,73 @@ set(CMAKE_CXX_STANDARD 20)
...
-  # Define preprocessor macro to indicate CCCL >= 3.0.0 is available
-  target_compile_definitions(cccl_interface INTERFACE NVMOLKIT_HAS_CCCL_GE_3=1)
-else()
-  if(CCCL_FOUND)
-    message(STATUS "CCCL ${CCCL_VERSION} found but version < 3.0.0")
+
+# Create the cccl_interface library
+add_library(cccl_interface INTERFACE)
+
+if(CCCL_FOUND)
```
scal444: Check out the versions here:

https://github.com/NVIDIA/cccl?tab=readme-ov-file#api-versioning

We can probably just get it from the header

src/minimizer/bfgs_minimize_permol_kernels.cu
scal444: Yeah I'll work on that

src/utils/cub_helpers.cuh
evasnow1992: I see [here](https://docs.nvidia.com/cuda/archive/12.9.0/cuda-toolkit-release-notes/index.html) cuda toolkit 12.9 is bundled with cccl 2.8.0, which still provides the old APIs?

CMakeLists.txt
evasnow1992: Yes, I was using cmake-format 0.6.13

src/utils/cub_helpers.cuh
evasnow1992: Yes, added its default value in cmake file and removed this check

src/utils/cub_helpers.cuh
evasnow1992: Moved this to butina.cu

CMakeLists.txt
evasnow1992: Yes, will use this instead of manual decoding. This also helps resolve the linking error with cccl >= 2.8.0 with the old version decoding approach.

src/butina.cu
```diff
@@ -292,7 +290,64 @@ __global__ void assignSingletonIdsKernel(const cuda::std::span<int> clusters, co
   }
 }
 
-constexpr int argMaxBlockSize = 512;
+#if CUB_VERSION >= 200800
+
+//! Wrapper for CUB's DeviceReduce::ArgMax (CCCL >= 2.8.0)
+//! Uses the new API that returns max value and index separately
+template <typename InputIteratorT>
+inline cudaError_t DeviceArgMax(void*          d_temp_storage,
```
scal444: inline isn't needed, that's only sometimes needed in headers.

src/butina.cu
```diff
@@ -292,7 +290,64 @@ __global__ void assignSingletonIdsKernel(const cuda::std::span<int> clusters, co
...
+                                size_t&        temp_storage_bytes,
+                                InputIteratorT d_in,
+                                int*           d_max_value_out,
+                                int*           d_max_index_out,
+                                int            num_items,
+                                cudaStream_t   stream = 0) {
+  ScopedNvtxRange range("CUB ArgMax (CCCL >= 2.8.0)");
```
scal444: Since this is async the range isn't useful here., it just returns immediately.

src/butina.cu
```diff
@@ -292,7 +290,64 @@ __global__ void assignSingletonIdsKernel(const cuda::std::span<int> clusters, co
...
+                                   temp_storage_bytes,
+                                   d_in,
+                                   d_max_value_out,
+                                   d_max_index_out,
+                                   static_cast<int64_t>(num_items),
+                                   stream);
+}
+
+//! Helper class to manage CUB temporary storage for ArgMax (CCCL >= 2.8.0)
+class ArgMaxTempStorage {
```
scal444: I'm not sure why this is called ArgMaxTempStorage when it actually does the argmax. 

I'd suggest naming the class ArgMaxRunner or something like that, and then include the old path within the class so that we don't need ifdefs everwhere, the ifdefs would just be in the class. Then in the loops, you'd instantiate an ArgMaxRunner called argMaxRunner or whatever.

src/butina.cu
```diff
@@ -292,7 +290,64 @@ __global__ void assignSingletonIdsKernel(const cuda::std::span<int> clusters, co
...
+    // Determine temporary storage requirements
+    DeviceArgMax(nullptr,
+                 temp_storage_bytes_,
+                 static_cast<int*>(nullptr),
+                 static_cast<int*>(nullptr),
+                 static_cast<int*>(nullptr),
+                 static_cast<int>(num_items),
+                 stream_);
+    // Allocate temporary storage
+    cudaCheckError(cudaMallocAsync(&temp_storage_, temp_storage_bytes_, stream_));
```
scal444: Can use AsyncDeviceVector<uint8_t> for this and wouldn't need the destructor.

src/butina.cu
evasnow1992: Removed it.

src/butina.cu
evasnow1992: Thank you for pointing this out. Removed this nvtx range.

src/butina.cu
evasnow1992: Good point. I was hesitant to wrap both paths into one class since the CUB path requires an additional initialization step for temporary storage (hence the original class name), and it results in multiple version checks inside the class compared to a single check in our old innerButinaLoop before the recent neighborlist changes. But I agree that encapsulating them now makes the code cleaner with our current setup. Updated the class name to ArgMaxRunner.

src/butina.cu
evasnow1992: Switched to AsyncDeviceVector<uint8_t>

scal444 (approved):
Looks good!

PR #68, opened 2025-12-11
Fix setup.py (on some python versions), and rework cuda targets

scal444:
We now target sm70 only if CUDA < 13.

The PTX version stored is now 90. SM 80 works for 86 and 89. Removed 89 since it's less common and not HPC, doesn't need to be all the way optimized. Kept 86 because torch does too.

Embed PTX for 90 now, so that > 90 has a more recent PTX. Set default to 80

Updated verification script for new versions


evasnow1992 (approved):
Thank you for bumping the version. LGTM.

PR #73, opened 2026-01-21
Substructure: Atom data and boolean tree + tests

scal444:
First PR for substructure search. Includes definitions for bit-packed atom data and comparison functions, and an implementation of a boolean tree walk. 

The boolean tree reference recursive bits that are not yet implemented, but can be tested with dummy data.

### DIFF
--- a/src/substruct/atom_data_packed.h
+++ b/src/substruct/atom_data_packed.h
@@ -1,0 +1,380 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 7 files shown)


src/substruct/atom_data_packed.h
```diff
@@ -0,0 +1,380 @@
...
+  static constexpr int kDegreeBits    = 6;     ///< Number of bits for degree field (max 63)
+  static constexpr int kDegreeMask    = 0x3F;  ///< Mask for 6-bit degree (bits 0-5)
+  static constexpr int kIsAromaticBit = 6;     ///< Bit position within degree byte for isAromatic
+  static constexpr int kIsInRingBit   = 7;     ///< Bit position within degree byte for isInRing
+  /// @}
+
+  /// @name Compacted 4-bit field masks and bit positions
+  /// @{
+  static constexpr int k4BitMask              = 0x0F;  ///< Mask for 4-bit fields (max value 15)
+  static constexpr int kNumRingsBits          = 0;     ///< Low 4 bits of byte 1
```
evasnow1992, line 106: I notice for several atom properties (numRings, ringBondCount, numImplicitHs, numHeteroatomNeighbors), we're using 4-bit fields with a max value of 15. Should we add bounds checking or saturation in the setters? Currently, values > 15 are silently truncated, which could cause unexpected behavior.

evasnow1992 (approved):
Thank you for adding packed atom data and boolean tree operations. Only one minor comment. Also may considering adding some edge case tests like for empty trees, and/or tests for comparison operations like GreaterThan. Not required though

src/substruct/atom_data_packed.h
scal444: I've got those checks in the code that uses this in one of the next PRs

PR #75, opened 2026-01-26
Add Substructure Molecule Batch definitions and helpers

scal444:
Molecule host and device classes as well as a device view for kernel execution.

Code to parse ROMols into the new mol format. 

A few extra things here that need to be in for definitions but will be expanded on in future PRs:
* Debug info control
* More query data structures and recursive data pattern defnitions
* Alternate bond representations for more efficent kernel traversal.


src/substruct/packed_bonds.h
```diff
@@ -0,0 +1,107 @@
...
+/**
+ * @brief Flags for bond query matching.
+ *
+ * Bonds can have queries like `-&!@` (single AND not ring bond).
+ */
+enum BondQueryFlags : uint8_t {
+  BondQueryNone             = 0,
+  BondQueryIsRingBond       = 1 << 0,  ///< Bond must be in a ring (@)
+  BondQueryNotRingBond      = 1 << 1,  ///< Bond must NOT be in a ring (!@)
+  BondQuerySingleOrAromatic = 1 << 2,  ///< SingleOrAromaticBond query (matches single or aromatic only)
```
evasnow1992, line 32: Just to double confirm, BondQuerySingleOrAromatic, BondQueryDoubleOrAromatic and BondQueryAromaticOnly are defined but not really used. I assume they will be handled in a future PR?

evasnow1992 (approved):
Thank you for adding packed bonds and molecular infra

src/substruct/packed_bonds.h
scal444: Yup! And For more context, about the specifics of that enum, the AI figured out that given the fewer things a bond can be, you can actually explicitly enumerate all the possible options that make any sense without having to do a full and/or/not boolean tree like we do with atoms.

PR #79, opened 2026-01-29
Add thread safe queue

scal444:
This will be used for several producer/consumer interfaces, including worker threads receiving from preprocessor threads, and an RDKit fallback path for molecules we can't handle.


evasnow1992 (approved):
Thank you for isolating this component out for review before following changes. Very clean implementation.

PR #80, opened 2026-01-29
Add substruct algorithms

scal444:


src/substruct/substruct_algos.cuh
```diff
@@ -0,0 +1,677 @@
...
+        if (pIdx == 0 && laneId == 0 && level <= 5) {
+          printf("[GSI] Level %d partial 0 mapping: ", level);
+          for (int q = 0; q < numQueryAtoms; ++q) {
+            printf("%d ", (int)partial[q]);
+          }
+          printf("\n");
+        }
+      }
+
+      for (int tBase = 0; tBase < numTargetAtoms; tBase += kWarpSize) {
```
evasnow1992, line 504: Should we also check earlyExitFlag inside this loop?

src/substruct/substruct_algos.cuh
```diff
@@ -0,0 +1,677 @@
...
+ * @param overflowA Pre-allocated overflow buffer A (ping)
+ * @param overflowB Pre-allocated overflow buffer B (pong)
+ * @param maxOverflow Entries per overflow buffer
+ * @param matchCount Output: number of matches found
+ * @param reportedCount Output: number of matches written (StoreMatches) or painted (PaintBits)
+ * @param matchIndices Output buffer (only used in StoreMatches mode)
+ * @param maxMatches Maximum matches to write (only used in StoreMatches mode)
+ * @param matchOffset Offset into output (only used in StoreMatches mode)
+ * @param paintParams Paint mode parameters (only used in PaintBits mode)
+ * @param maxMatchesToFind Stop searching after this many matches (-1 = no limit)
```
evasnow1992: If I understand the algorithm correctly, I think there is a chance that one thread sets earlyExitFlag = 1, other warps may have already found additional matches. In that case the final matchCount can exceed maxMatchesToFind?

evasnow1992 (approved):
Two clarification questions. Changes look good to me.

src/substruct/substruct_algos.cuh
scal444: There's a second check before write, so I don't think that happened. But you bring up a good point, there's a few things going on here:

1. There was technically a data race, because the earlyExitFlag wasn't atomic. It was benign
2. I think there's a tradeoff between constantly reading from shared memory for the exit condition and having the work savings from the early exit. I'm not sure the benefit each way
3. This all ever happens at the final depth, since it's a BFS, and a match cna't be written until you've reached the final atom and built the partial up all the way. I think we should rework to optimize this. I've added a TODO, and we can revisit the early exit logic with microbenchmarks then. 

For now, I've removed the early exit flag and have the inner if conditional check its own match idx only. I also renamed maxMatches to indiccate it's a storage limit rather than the other parameter maxMatchesToFind

src/substruct/substruct_algos.cuh
```diff
@@ -0,0 +1,677 @@
...
+      ++candIdx;
+    }
+
+    if (!foundCandidate) {
+      // Backtrack
+      state.candidateIdx[state.depth] = 0;
+      --state.depth;
+      if (state.depth > 0) {
+        ++state.candidateIdx[state.depth];
+      } else if (state.depth == 0) {
```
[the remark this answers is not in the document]
scal444: Done

src/substruct/substruct_algos.cuh
scal444: See other comment, but good point

PR #81, opened 2026-01-30
Implement substructure search with full recursion

scal444:


src/substruct/substruct_constants.h
```diff
@@ -0,0 +1,33 @@
...
+
+//! Maximum size for GPU substruct path
+constexpr int kMaxTargetAtoms = 128;
+constexpr int kMaxQueryAtoms  = 64;
+
+/// Entries per overflow buffer per block
+constexpr int kOverflowEntriesPerBuffer = 2048;
+
+/// Maximum nesting depth for recursive SMARTS patterns
+constexpr int kMaxSmartsNestingDepth = 4;
```
evasnow1992, line 29: Just curious, does RDKit also assume a maximum nesting depth of 4? If not, what happens if the nesting depth exceeds 4 in our implementation? Fall back to calling RDKit?

src/substruct/substruct_kernels.cu
```diff
@@ -0,0 +1,1098 @@
...
+                                                                   results.matchIndices,
+                                                                   maxMatches,
+                                                                   matchOffset,
+                                                                   maxMatchesToFind,
+                                                                   countOnly);
+    }
+
+  } else if constexpr (Algo == SubstructAlgorithm::GSI) {
+    constexpr int kBlockSizeT   = getBlockSizeForConfig<MaxTargetAtoms>();
+    constexpr int kMaxPartialsT = getMaxPartialsForSM<MaxTargetAtoms, MaxQueryAtoms>(86, kBlockSizeT);
```
evasnow1992: This may be trivial, but why we choose to hard-code 86?

src/substruct/substruct_search_internal.cpp
```diff
@@ -0,0 +1,415 @@
...
+    const int16_t* src = hostBuffer.matchIndices.data() + u.miniBatchLocalOffset;
+
+    for (int m = 0; m < u.reportedMatches; ++m) {
+      auto& match = targetMatches.emplace_back(u.queryAtoms);
+      for (int a = 0; a < u.queryAtoms; ++a) {
+        match[a] = src[m * u.queryAtoms + a];
+      }
+    }
+  }
+}
```
evasnow1992, line 379: Just to clarify, here the creation of matchRefs is inside the lock, but the writes to targetMatches are outside the lock. Will there be a race condition concern?"

tests/test_substruct_search.cu
```diff
@@ -0,0 +1,1935 @@
...
+        }
+      }
+    }
+  }
+};
+
+// Instantiate parameterized tests for all algorithms
+INSTANTIATE_TEST_SUITE_P(AllAlgorithms,
+                         SubstructureSearchTest,
+                         ::testing::Values(SubstructAlgorithm::GSI),
```
evasnow1992, line 209: Double confirm here if we only want to test GSI but not VF2. I know VF2 is mostly used internally as reference, but the use of `AllAlgorithms` here can be a little confusing.

evasnow1992 (approved):
Thank for you putting all these together. A few clarification questions.

src/substruct/substruct_search_internal.cpp
scal444: Each vector has a separate allocation - so we're locking when modifying the internal allocation, but then matchRefs[i] points to a different location in memory and can safely be written to, after matchRefs[i] is instantiated. 

If you think of a vector as a trio of `{ptr, size, capacity}`, matchrefs.push_back() populates a bunch of these back to back
`{ ptr size cap}, {ptr size cap}`

Then outside of the lock, we're effectively accessing `ptr` which doesn't affect the rest of the vector.

src/substruct/substruct_constants.h
scal444: Good question, right now it throws. But we should handle it like the other items with fallback instead. Just changed and added a unit test!

src/substruct/substruct_kernels.cu
scal444: Fixed! This should be based on cuda arch - 86 was for cc 86, my laptop

tests/test_substruct_search.cu
scal444: Yes, thank you! Added them back, VF2 supports the nonrecursive cases. Split off the recursive cases so that they're no longer in "All algorithms" since it's just GSI

PR #83, opened 2026-02-04
Add Substruct Integration tests

scal444:
Integration test data drawn from both RDKit and external sources. 

Data was processed using scripts, separated into

* supported by us
* Not supported by us
* not supported by rdkit (invalid, for our purposes)

with the original file kept for reference. Since we support most features, most not supporteds (but not all) are empty


tests/smarts_filter.cpp
```diff
@@ -0,0 +1,206 @@
...
+/**
+ * @brief Extract the first whitespace-delimited token from a line.
+ *
+ * Handles formats like:
+ *   - "SMARTS" (just the pattern)
+ *   - "SMARTS   description" (pattern + description)
+ *   - "SMARTS   description   # comment" (pattern + description + comment)
+ */
+std::string extractSmarts(const std::string& line) {
+  const char* ws    = " \t\n\r";
```
evasnow1992: These file preprocessing steps are very similar to the file reading functions defined in mol_data.h. I know they're not identical, but they share a similar parsing logic, so we might consider consolidating to reduce code duplication.

tests/test_substruct_integration.cu
```diff
@@ -0,0 +1,658 @@
...
+using nvMolKit::printValidationResultDetailed;
+using nvMolKit::ScopedStream;
+using nvMolKit::SubstructAlgorithm;
+using nvMolKit::SubstructSearchConfig;
+using nvMolKit::SubstructSearchResults;
+using nvMolKit::validateAgainstRDKit;
+using nvMolKit::testing::readSmartsFileWithStrings;
+using nvMolKit::testing::readSmilesFileWithStrings;
+
+std::vector<const RDKit::ROMol*> getRawPtrs(const std::vector<std::unique_ptr<RDKit::ROMol>>& mols) {
```
evasnow1992: Looks like the same functionality as `makeMolsView` defined in mol_data.h?

tests/test_substruct_integration.cu
```diff
@@ -0,0 +1,658 @@
...
+  int64_t multiGpuTotal  = 0;
+  int     mismatches     = 0;
+
+  for (int t = 0; t < singleGpuResults.numTargets; ++t) {
+    for (int q = 0; q < singleGpuResults.numQueries; ++q) {
+      const int singleCount = singleGpuResults.matchCount(t, q);
+      const int multiCount  = multiGpuResults.matchCount(t, q);
+      singleGpuTotal += singleCount;
+      multiGpuTotal += multiCount;
+      if (singleCount != multiCount) {
```
evasnow1992, line 592: Should we also check exact mappings here? I see we use validateAgainstRDKit for comparing with rdkit. Wonder if that function can be reused (or the `matchSetsEqual` function called inside it) for comparing results between single and multiple GPUs

tests/test_substruct_label_integration.cu
```diff
@@ -0,0 +1,358 @@
...
+                              targetSmiles,
+                              querySmarts,
+                              "false positive",
+                              [&](int t, int q) -> std::pair<int, int> {
+                                auto it = fpInfo.find({t, q});
+                                return it != fpInfo.end() ? it->second : std::pair{0, 0};
+                              });
+  }
+
+  EXPECT_EQ(totalFalsePositives, 0)
```
evasnow1992, line 414: We are not asserting on false negatives nor report them. Is there a reason we are more tolerant towards false negatives?

evasnow1992 (approved):
Thank you for adding integration tests. A few minor questions.

tests/test_substruct_integration.cu
```diff
@@ -0,0 +1,658 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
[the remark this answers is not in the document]
scal444: Done

tests/test_substruct_integration.cu
scal444: Done

tests/test_substruct_label_integration.cu
scal444: Ah this was a brute force suppression because recursive smarts aren't supported in this workflow. Changed to do the False negative check in the case of nonrecursive SMARTS

tests/test_substruct_integration.cu
scal444: Done

tests/smarts_filter.cpp
scal444: Consolidated some of the logic

PR #85, opened 2026-02-05
Python interface and tests for substruct integration + Benchmark

scal444:


nvmolkit/tests/test_substructure.py
```diff
@@ -0,0 +1,1048 @@
...
+
+        config.uniquify = True
+        assert config.uniquify is True
+
+        config.uniquify = False
+        assert config.uniquify is False
+
+
+# =============================================================================
+# hasSubstructMatch Tests
```
evasnow1992, line 568: This test script tests both `getSubstructMatches` and `hasSubstructMatch`, but not `countSubstructMatches` explicitly. Should we add a few tests for it?

evasnow1992 (approved):
Thank you for adding the python wrappers and benchmark script. Only one comment regarding adding more tests for countSubstructMatches. Changes look good to me.

nvmolkit/tests/test_substructure.py
scal444: Good catch, done!

PR #86, opened 2026-02-08
Consolidate query preprocessing to 1 time, rather than per minibatch

scal444:
A part of query prep was done repeatedly, now is just once at startup.


evasnow1992 (approved):
LGTM.

PR #87, opened 2026-02-08
Remove cudaMemSetAsync calls by fusing with initial kernels.

scal444:
cudamemsetAsync was taking nontrivial CPU time to dispatch, and is part of the bottleneck for tight loops where both preprocessing and compute kernels are fast (such as large query datasets of simple queries). 

This moves the initial resetting into the first called kernel per minibatch and reduces the number of API calls.

### DIFF
--- a/src/substruct/substruct_kernels.cu
+++ b/src/substruct/substruct_kernels.cu
@@ -250,23 +250,20 @@ __global__ void labelMatrixPaintKernelT(TargetMoleculesDeviceView  targets,
                                         int                        firstTargetIdx,
                                         const uint32_t*            recursiveMatchBits,
                                         int                        maxTargetAtoms,
-                                        uint32_t*                  recursiveBitsToZero,
-                                        int                        recursiveBitsSizeToZero,
-                                        uint8_t*                   overflowFlagsToZero,
-                                        int                        overflowFlagsSizeToZero) {
+                                        ZeroBuffersSpec            zeroBuffers) {
   // Zero buffers using grid-stride loop (only on first label kernel launch)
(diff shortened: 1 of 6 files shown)


src/substruct/substruct_kernels.cu
```diff
@@ -432,6 +452,9 @@ __global__ void substructMatchKernelT(TargetMoleculesDeviceView
     __shared__ PartialMatchT<MaxQueryAtoms> gsiPartials[kMaxPartialsT * 2];
 
     uint8_t* pairOverflowFlag = results.overflowFlags ? &results.overflowFlags[miniBatchIdx] : nullptr;
+    if (pairOverflowFlag != nullptr && threadIdx.x == 0) {
+      *pairOverflowFlag = 0;
+    }
```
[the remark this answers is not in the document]
scal444: No, only thread 0 ever writes as well as reads, so no possible race.

src/substruct/substruct_kernels.h
```diff
@@ -95,6 +99,10 @@ void launchLabelMatrixPaintKernel(SubstructTemplateConfig    config,
                                   int                        firstTargetIdx,
                                   const uint32_t*            recursiveMatchBits,
                                   int                        maxTargetAtoms,
+                                  uint32_t*                  recursiveBitsToZero,
```
evasnow1992: Since these four newly added parameters are only used in the first call, might consider wrapping them into a optional struct parameter. Just a minor API design suggestion, not required.

evasnow1992 (approved):
LGTM

src/substruct/substruct_kernels.h
scal444: Good thought, done!

PR #90, opened 2026-02-12
Send substruct algo test buffers to device explicitly

scal444:
I think these were accidentally being handled via [HMM](https://developer.nvidia.com/blog/simplifying-gpu-application-development-with-heterogeneous-memory-management/) which is why the test wasn't crashing before.


evasnow1992 (approved):
Thank you for fixing this!

PR #91, opened 2026-02-12
Fix -Werror warnings with gcc 13 and CUDA 13.1

evasnow1992:
Address compilation warnings promoted to errors by -Werror:

- nvtx.h: Use {} instead of {0} for nvtxEventAttributes_t initialization to fix -Wmissing-field-initializers
- molecules.cpp: Comment out unused parameter parentAtomIdx to fix -Wunused-parameter
- substruct_search_internal.h: Change forward declaration from class to struct to match definition, fixing -Wmismatched-tags
- test_boolean_tree.cu: Remove unused functions makeAtomWithNumRings and makeAtomWithTotalValence (NVCC error #177)
- test_flat_bit_vect_device.cu: Change int loop variables to size_t to match constexpr std::size_t bounds, fixing -Wsign-compare
- test_graph_labeler_recursive.cu: Comment out unused parameter targetHost to fix -Wunused-parameter
- test_substruct_integration.cu: Add explicit .executorsPerRunner and .gpuIds fields to designated initializers to fix -Wmissing-field-initializers
- test_substruct_search.cu: Change const std::string& to const auto& in range-for over initializer_list to fix -Wrange-loop-construct

### DIFF
--- a/src/substruct/molecules.cpp
+++ b/src/substruct/molecules.cpp
@@ -1832,9 +1832,9 @@ int collectPatternsFromMolecule(const RDKit::ROMol*                 mol,
                                 std::vector<RecursivePatternEntry>& patterns,
                                 int&                                nextPatternId,
                                 int                                 parentIdx,
-                                int                                 parentAtomIdx,
-                                int                                 currentDepth,
-                                int&                                nextLocalId) {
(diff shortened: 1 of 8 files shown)


scal444 (approved):
Thanks!

PR #94, opened 2026-02-17
Update docs and python interfaces for v0.4

scal444:
* Added wrappers around C++ to better render docstrings in python
* Also added missing v0.3 stuff. It had been hosted but not merged into main


evasnow1992 (approved):
LGTM.

PR #98, opened 2026-02-20
Fix shared memory overflow due to config setting error

scal444:
If a recursive pattern is larger than the base pattern, we still choose configs based on the base pattern, leading to potential OOB in shared memory. The change is to use the largest of {base, all recursive patterns}


evasnow1992 (approved):
Thank you for fixing this memory issue with long recursive patterns.

PR #113, opened 2026-03-20
Fix empty results in uniquify if all results were unique to start with

scal444:
There was an invalid guard which would cause us to not return the results.

Most of the diff is expanding our integration tests to cover this case. New tests failed before fix.

Addresses #112


evasnow1992 (approved):
Thank you for identifying and fix this issue. Code changes look good to me.

PR #122, opened 2026-03-30
Add forcefield constraints. Generic but right now just applied to MMFF

scal444:
Constraints are generic but right now just applied to MMFF. 

Improved python docstrings and explained how to access individual batch members to apply constraints, get energies, and gradients.


evasnow1992 (approved):
Thank you for adding MMFF forcefield constraints. The two newly added files, forcefield_constraints.h/.cpp are missing the copyright headers. Other changes look good for me.

Issue #70, opened 2025-12-28
Forcefield variants and properties addition


ethanev:
Hi, would it be possible to add functionality to use variants of MMFF and alter properties of the forcefield (I believe this would be implementing the Forcefield module)? I commonly need to optimize structures in different dielectrics and get energies like below (cids is just a list of ints for the conformers previously generated and mol is a molecule with generated conformers):

`   
dielectric = 80.0
mp = AllChem.MMFFGetMoleculeProperties(mol, mmffVariant='MMFF94s')
mp.SetMMFFDielectricConstant(dielConst=dielectric)
ff = AllChem.MMFFGetMoleculeForceField(mol, mp)
AllChem.OptimizeMoleculeConfs(mol, ff, numThreads=num_threads)
    
for cid in cids:
    ff = AllChem.MMFFGetMoleculeForceField(mol, mp, confId=cid)
    e = ff.CalcEnergy()
`

Would it be possible to add such functionality? This would significantly impact my discovery workflow!
Thanks,
Ethan

scal444:
Thanks for the feature request! It's something we could do some refactoring, similar to issue #26. It isn't our top priority this cycle but will discuss with the team since we have 2 requests for a FF module.

scal444:
Support for this is merged and will be released in v0.5 in May. Details [here](https://github.com/NVIDIA-Digital-Bio/nvMolKit/blob/main/docs/forcefield.rst), and our hosted docs will have this page when we release. 

Please reopen if there's any missing features you need

PR #124, opened 2026-03-31
Add custom options for MMFF optimization

scal444:
Allows better control over dielectric, interfrag interactions, and nonbond interaction limit. Partial implementation of #70

### DIFF
--- a/nvmolkit/mmffOptimization.cpp
+++ b/nvmolkit/mmffOptimization.cpp
@@ -75,9 +75,8 @@ BOOST_PYTHON_MODULE(_mmffOptimization) {
     "MMFFOptimizeMoleculesConfs",
     +[](const boost::python::list&            molecules,
         int                                   maxIters,
-        const boost::python::object&          propertiesObj,
+        const boost::python::list&            propertiesList,
         const nvMolKit::BatchHardwareOptions& hardwareOptions) -> boost::python::list {
-      // Convert Python list to std::vector<RDKit::ROMol*>
       std::vector<RDKit::ROMol*> molsVec;
       molsVec.reserve(len(molecules));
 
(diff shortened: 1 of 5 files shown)


nvmolkit/mmffOptimization.py
```diff
@@ -117,8 +127,39 @@ def MMFFOptimizeMoleculesConfs(
...
+    interfrag_flags = _normalize_scalar_or_list(ignoreInterfragInteractions, "ignoreInterfragInteractions")
+    native_properties = [
+        make_internal_mmff_properties(
+            props,
+            non_bonded_threshold=float(threshold),
+            ignore_interfrag_interactions=bool(ignore_interfrag),
+        )
+        for props, threshold, ignore_interfrag in zip(properties_list, thresholds, interfrag_flags)
+    ]
+    if len(native_properties) == 1:
```
evasnow1992: Is there any specific reason we want to extract this single property instead of just passing in it as a list of one element?

nvmolkit/mmffOptimization.cpp
```diff
@@ -55,23 +89,29 @@ BOOST_PYTHON_MODULE(_mmffOptimization) {
         molsVec.push_back(mol);
       }
 
-      nvMolKit::MMFFProperties properties;
-      properties.nonBondedThreshold = nonBondedThreshold;
-      auto result = nvMolKit::MMFF::MMFFOptimizeMoleculesConfsBfgs(molsVec, maxIters, properties, hardwareOptions);
+      std::vector<std::vector<double>>            result;
+      boost::python::extract<boost::python::list> propertiesListExtract(propertiesObj);
+      if (propertiesListExtract.check()) {
```
evasnow1992: Related to my comment below. We are allowing the python API to pass in either a scalar or a list, which makes the C++ side here having to handle both cases. Should we just always pass in a list instead?

nvmolkit/mmffOptimization.py
```diff
@@ -19,21 +19,26 @@
 optimization for multiple molecules and conformers using CUDA and OpenMP.
 """
 
+from collections.abc import Sequence
 from typing import TYPE_CHECKING
 
-from rdkit.Chem import AllChem
+from rdkit.Chem import AllChem, rdForceFieldHelpers
```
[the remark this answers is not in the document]
scal444: Done

nvmolkit/mmffOptimization.py
scal444: It matters for which src cpp path we take - https://github.com/NVIDIA-Digital-Bio/nvMolKit/blob/main/src/minimizer/bfgs_mmff.h#L38-L44, but you're right we didn't need it at this level. Fixed

evasnow1992 (approved):

PR #133, opened 2026-04-21
Mmff minimize hardware options

scal444:
Extends the original MMFF minimization interface to support constraints, then the new forcefield API wraps the extended interface, allowing multi-GPU and batch options without further code duplication

### DIFF
--- a/nvmolkit/batchedForcefield.cpp
+++ b/nvmolkit/batchedForcefield.cpp
@@ -35,13 +35,6 @@ namespace bp = boost::python;
 
 namespace {
 
-struct ConformerEntry {
-  RDKit::ROMol*     mol;
-  int               molIdx;
-  RDKit::Conformer* conformer;
-  uint32_t          atomStart;
-};
-
 std::vector<std::vector<double>> splitGradients(const std::vector<double>& flatGrad,
                                                 const std::vector<int>&    atomStarts,
                                                 int                        dim) {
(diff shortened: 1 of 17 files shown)


nvmolkit/batchedForcefield.cpp
```diff
@@ -253,26 +210,70 @@ static std::vector<PerMolConstraints> extractAllConstraints(const bp::list& dist
...
+
+  bp::tuple minimize(int maxIters, double gradTol) {
+    auto result =
+      nvMolKit::MMFF::MMFFMinimizeMoleculesConfs(mols_, maxIters, gradTol, properties_, constraints_, hwOpts_);
+
+    dirty_ = true;
+
+    bp::list energies;
+    bp::list converged;
+    for (size_t molIdx = 0; molIdx < result.energies.size(); ++molIdx) {
```
evasnow1992: Optional. Might be able to reuse reshapeToNested here instead of the nested loops?

nvmolkit/batchedForcefield.cpp
```diff
@@ -295,48 +294,26 @@ class NativeMMFFBatchedForcefield {
...
-  }
-
-  bp::list computeGradients() {
-    gradDevice_.zero();
-    throwIfCudaError(forcefield_->computeGradients(gradDevice_.data(), positionsDevice_.data()), "computeGradients");
-    auto perSystem = splitGradients(copyDeviceVector(gradDevice_), forcefield_->atomStartsHost(), 3);
-    return reshapeGradientsToNested(perSystem, numConformersPerMol_);
+  void rebuildIfDirty() {
+    if (dirty_) {
+      buildForcefield();
```
evasnow1992: I may have missed some details here, but do we really need a full reconstruction? minimize() should only have updated positions — those are stale and need to be synced before computing energy or gradients, but the forcefield parameters haven't changed, so a full rebuild seems unnecessary. Not a blocker, though it might impact performance.

evasnow1992 (approved):
One clarification question regarding the requirement of a full rebuild for computing energy and gradient. Not a blocker. Changes look good to me.

nvmolkit/batchedForcefield.cpp
scal444: Not directly, but consolidated usages to a single template.

nvmolkit/batchedForcefield.cpp
scal444: You were right thanks, changed to just update positions - this let me remove a few now unused data structures too.

PR #135, opened 2026-04-23
TFD kernels: resolve divergence investigation TODOs with ncu findings

evasnow1992:
Replace two "maybe profile this later" TODOs on the TFD GPU kernels with the concrete ncu measurements that closed each question.
- dihedralKernel's findMolecule binary search: ncu reports branch efficiency >=99.95%. Consecutive work items are molecule-contiguous, so the binary search does not cause meaningful warp divergence. A lookup-table replacement is not needed.
- tfdMatrixKernel's torsion-type switch: ncu reports branch efficiency >=99.93%. All threads in a block iterate the torsion list in lockstep, so the Single/Ring/Symmetric branch produces no SIMT divergence. A host-side torsion sort by type is not needed.

### DIFF
--- a/src/tfd/tfd_kernels.cu
+++ b/src/tfd/tfd_kernels.cu
@@ -42,9 +42,6 @@ __device__ __forceinline__ int findMolecule(const int* starts, int numMolecules,
 //! One thread per (conformer, quartet) work item.
 //! Uses binary search on dihedralWorkStarts to find the molecule,
 //! then computes (confIdx, quartetIdx, outIdx) arithmetically.
-//! TODO: The binary search may cause warp divergence for large molecule counts.
-//! Consider precomputing a flat workItem→moleculeIdx lookup table if ncu profiling
-//! shows this is a bottleneck.
 __global__ void dihedralKernel(const int totalWorkItems,
                                const float* __restrict__ positions,
                                const int* __restrict__ confPositionStarts,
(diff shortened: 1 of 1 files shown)


src/tfd/tfd_kernels.cu
```diff
@@ -42,9 +42,9 @@ __device__ __forceinline__ int findMolecule(const int* starts, int numMolecules,
 //! One thread per (conformer, quartet) work item.
 //! Uses binary search on dihedralWorkStarts to find the molecule,
 //! then computes (confIdx, quartetIdx, outIdx) arithmetically.
-//! TODO: The binary search may cause warp divergence for large molecule counts.
-//! Consider precomputing a flat workItem→moleculeIdx lookup table if ncu profiling
-//! shows this is a bottleneck.
+//! ncu-investigated (benchmark_smiles.csv, 50 conformers, 2026-04-23): branch
+//! efficiency >=99.95%. Consecutive work items are molecule-contiguous, so the
```
scal444: I'd rather not have historical comments like this, "no change needed" is an LLM response to a question, not a comment about the current state of the code.

scal444: I've got a rule in cursor about the models not writing this sort of comment, but they do it anyway sometimes

evasnow1992: Will simply remove the old TODO comments

scal444 (approved):

PR #143, opened 2026-04-27
Sample SMILES/pickle inputs randomly in substruct benchmark

scal444:
`benchmarks/substruct_bench.py` previously took the first `--num_mols` entries from the top of the SMILES or pickle input. Switch to a uniform random sample so benchmark inputs reflect a representative slice of large datasets. Critical for files sorted by molecule size.

For the smiles case, builds up the test set in place, so we iterate the whole file but don't instantiate all the SMILES at once. For pickles, we do take that memory hit for now.

### DIFF
--- a/benchmarks/substruct_bench.py
+++ b/benchmarks/substruct_bench.py
@@ -55,6 +55,7 @@ Usage:
 import argparse
 import gc
 import pickle
+import random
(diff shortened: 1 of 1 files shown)


evasnow1992 (approved):

Issue #42, opened 2025-11-03
Clean up CMake Include Directories


scal444:
Having each library link separately specify its includes doesn't serve any function, and the header paths have relevant info sometimes.

scal444:
Finished in https://github.com/NVIDIA-BioNeMo/nvMolKit/pull/191

PR #146, opened 2026-04-28
CMake cleanup 1

scal444:
Part 1 towards Issue #42

- Declare NVMOLKIT_BUILD_PYTHON_BINDINGS as a real option(); previously it was used in five files but never declared.
- Replace the unconditional set(CMAKE_BUILD_TYPE ... FORCE) with a guarded default that only fires for single-config generators when no build type has been chosen, and add the supported types as cache STRINGS metadata.
- Replace ${CMAKE_SOURCE_DIR} usage in target_include_directories with ${PROJECT_SOURCE_DIR} (or remove redundant entries), so the project is safe to consume as a sub-add and does not leak its source-tree assumption.
- Centralize cmake_policy(SET CMP0167 NEW) at the top of the root CMakeLists.txt; remove duplicates from cmake/boost.cmake and cmake/rdkit.cmake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,6 +12,9 @@
 
 cmake_minimum_required(VERSION 3.30)
 
+# Use the FindBoost replacement provided by the upstream Boost project.
+cmake_policy(SET CMP0167 NEW)
+
 project(
   nvmolkit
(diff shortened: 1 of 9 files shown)


evasnow1992 (approved):

Issue #140, opened 2026-04-27
Device input/return types for MMFF optimization and Conformer generation


scal444:
We currently re-pack all coordinates from ETKDG/MMFF/UFF back into the mol structures. A lot of workflows want to use coordinates in downstream GPU work, so they have to re-unpack and put data back on the device. It's also a slowdown between ETKDG->minimization workloads.

We should provide output APIs that are on-device for all of these, and input APIs for minimization for better chaining.

scal444:
The big question is what format(s) we should support. 

Options:

1. list/vector of [tensor], where tensor is (B, N_atoms, 3), with one tensor per molecule type.

2. Complete CSR. The tricky thing here is variable conformers per molecule, we'd need multiple levels of indexing.

3. Single padded tensor with masking (maybe multiple masks). Dims (N_mol, B_max, N_atoms_max, 3) 


We could support all of these natively, or just support the CSR with python side conversion functions

scal444:
Done for outputs.

The remaining case not currently handled is passing ETKDG results directly to MMFF or UFF on device. The problem is that we do host-side forcefield param construction relying on conformer coordinates. This gets very messy, and would require us to probably do forcefield assignment on the GPU.

Can reopen if there's a glaring need for it, such as if it's identified to be a major workflow blocker

PR #156, opened 2026-05-04
ETKDG device side output

scal444:
For now we're disabling pruning in this path, until we have GPU-side pruning. 

Includes some common functionality that will be used for device-side MMFF / UFF in a followup


scal444:
Partial support for #140

evasnow1992 (approved):

Issue #141, opened 2026-04-27
HW options Auto-tuning for large scale conformer work


scal444:
Performance is very hardware-option dependent, and hardware options are very dependent on the data types and the actual underlying hardware. We could wrap a quick Optuna optimization cycle with optuna as an optional install dependency

PR #157, opened 2026-05-04
Add autotuning functions for our main APIs

scal444:
Also fixed a bug introduced between rebases expecting Keyerror instead of valueerror


scal444:
KeyError bug would have been caught with python builds in CI, I'll add one in a different change

scal444:
This is PR 3 out of 5 or 6 for autotuning #141, we still need to plumb the autotuning to our actual benchmark scripts, add substruct config searches, and add documentation

evasnow1992 (commented):
All the four tune_*.py files are quite similar. May consider further consolidation in a follow-up PR. Not a blocker.

scal444:
> All the four tune_*.py files are quite similar. 

There are a few subtle differences, but agreed there's some consolidation that could be made. I'll revisit that with the substruct tune branch once they're both in

evasnow1992 (approved):
Changes look good to me. May consider consolidating the tune files in a follow-up PR.

PR #168, opened 2026-05-08
Template per-mol MMFF/UFF kernels on constraints

scal444:
The per-molecule MMFF and UFF energy/gradient device functions iterate over four optional constraint term types (distance, position, angle, torsion). When constraints are absent the loops still bring extra register pressure into the BFGS per-mol kernel, lowering occupancy.

Add a HasConstraints template parameter to MMFF::molEnergy / molGrad, UFF::molEnergy / molGrad, and the bfgsMinimizeKernel that drives the MMFF per-mol minimize. With if constexpr the constraint loops are compiled out of the no-constraint specialization. At launch time we inspect the host-visible sizes of the constraint device vectors via a new batchHasConstraints() helper and dispatch to the matching specialization. ETK and DG keep a single specialization since they do not consume MMFF-style constraint terms.


scal444:
@evasnow1992 this diff is mostly formatting

evasnow1992 (approved):

Issue #169, opened 2026-05-08
Integer overflow for large substruct worfklows


scal444:

scal444:
Fixed with #170

PR #170, opened 2026-05-08
Fix int32 overflow in substruct pair indexing

scal444:
For batches large enough that numTargets * numQueries exceeds INT32_MAX (e.g. 10M targets x 480 queries = 4.8B pairs), the flat pair index computed as `targetIdx * numQueries + queryIdx` wrapped to a negative int. That index was then sign-extended to size_t when used to subscript HasSubstructMatchResults::hasMatch and the counts vector, causing out-of-bounds writes and segfaults in hasSubstructMatch and countSubstructMatches. The same overflow corrupted unordered_map keys in SubstructSearchResults::matches.

Widen pairIndex computations and the map key type to int64_t. Cast to size_t at indexing sites. Kernel-side pair indices are unchanged because they operate in batch-local target coordinates bounded by batchSize.

### DIFF
--- a/src/substruct/substruct_results.h
+++ b/src/substruct/substruct_results.h
@@ -66,7 +66,7 @@ struct SubstructSearchConfig {
 struct SubstructSearchResults {
   /// Sparse storage: pairIndex -> vector of matches
   /// Each match is a vector<int> of target atom indices (one per query atom)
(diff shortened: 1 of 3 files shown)


scal444:
Fixes #169

evasnow1992 (approved):
Changes look good to me. Thank you for the clarification on keeping using int32 on the kernel side.

PR #171, opened 2026-05-11
Ship sm_103 SASS and fix cross-similarity dispatch on B300

scal444:
On B300 the host's supportsTensorOps fell through with no NVMOLKIT_CUDA_CC_103 macro and picked the SIMT tile shape, while the device-side __CUDA_ARCH__ >= 800 guard still selected the BMMA kernel, producing NaN outputs from the cross-similarity kernels. Adds 103-real to the CUDA >= 12.9 arch list, defines NVMOLKIT_CUDA_CC_103 in both detect paths, and teaches supportsTensorOps about sm_103.


evasnow1992 (approved):
LGTM. Thanks!

PR #172, opened 2026-05-13
RDKit 2026.3 support and minor bug fix

scal444:
* Removed system from boost targets. It's been header only since 1.67, so the target isn't needed as long as the include dir is right. Boost 1.90 (rdkit links to this) dropped system from their targets so it was a compile issue.

* Fixed a stream ordering bug in MMFF. It was latent, I think our recent constraint code caused it to be hit, but did not cause the issue directly.

* Track RDKit version to match results for multiple bug fixes.

* Remove flaky memory test. Due to memory pooling/caching, it was never reliably correct.


evasnow1992 (approved):
LGTM

PR #174, opened 2026-05-15
CMake and linking fixes for PyPi builds

scal444:
The pip-build path was globbing every .so in rdkit.libs/ into RDKit_LIBS and every libboost_* into Boost_LIBRARIES, then linking each into every nvmolkit Python module. That dragged libcairo and libquadmath onto each module's NEEDED list. libcairo in turn NEEDEDs libXrender/libX11/libXext; those are on the manylinux lib_whitelist so rdkit-pypi's auditwheel pass legally left them external, but the nvidia/cuda runtime container we test in doesn't ship them, so `import nvmolkit.fingerprints` failed at load.

Narrow RDKit_LIBS to the 16 components nvmolkit actually uses (mirroring the conda path's explicit list) and Boost_LIBRARIES to the same 4 (boost serialization / iostreams / python<ver> / numpy<ver>). Also patchelf with --force-rpath so the entry-point modules get DT_RPATH instead of DT_RUNPATH. The libs inside rdkit.libs/ have no rpath of their own and rely on RPATH inheritance to resolve second-level deps; rdkit's own python bindings do the same. Drop the unused ${RDKit_LIBS} link from _arrayHelpers, which touches zero RDKit symbols.

### DIFF
--- a/admin/distribute/repair_wheel.sh
+++ b/admin/distribute/repair_wheel.sh
@@ -86,10 +86,14 @@ unzip -q "${REPAIRED_WHEEL}" -d "${WORK}"
 
 # Each nvmolkit/_*.so resolves siblings in nvmolkit.libs/ (auditwheel's vendor
 # dir), and externally-shipped libs in rdkit.libs/ and nvidia/cuda_runtime/lib
-# under the same site-packages root.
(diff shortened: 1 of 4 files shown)


evasnow1992 (approved):

PR #175, opened 2026-05-18
Add downloadable agent skill and unit test for snippets

scal444:


evasnow1992 (approved):
Thank you for adding this skill. Changes look good to me!

PR #177, opened 2026-05-19
Perf & correctness fixes: RMSD warp-shuffle, TFD precision, eigensolver persistence, similarity cache, BFGS sign fix, fused Butina sync reduction

mooreneural:
### Correctness fixes

- **tfd_kernels.cu** Use double-precision sqrt for TFD pair-index decoding (float32 loses bits past ~1415 conformers); guard ring-torsion branch against numQ==0 division by zero.
- **bfgs_minimize.cu** Convergence denominator kept as RDKit's signed-energy form to preserve reference behavior. Added TODO to file upstream RDKit bug and gate the fix on RDKit version once merged (matching the pattern already used in scaleGradKernel).
- **triangle_smooth.cu** Replace cudaDeviceSynchronize() with cudaStreamSynchronize(data_.stream()) in copyToHost, scoping the sync to the correct stream.

### Performance fixes

- **conformer_rmsd.cu** Replace 17 sequential cub::BlockReduce + __syncthreads() calls in computePairRmsd with warp-shuffle reductions. Sync count drops 17→3 (Kabsch) or 17→1 (prealigned); shared memory drops ~1120→400 bytes per block.
- **symmetric_eigensolver.cu** Move AsyncDeviceVector<curandState> from a per-call heap allocation to a persistent states_ member of BatchedEigenSolver::Impl; reused across calls, grows only when batch size increases.
- **similarity_kernels.cu** Add per-device int8_t cache for tensor-op capability; cudaGetDeviceProperties called at most once per device per process instead of on every similarity launch.
- **clustering.py (fused_butina)** Reduce GPU↔CPU syncs from ~6 to 2 per iteration by batching neigh max+argmax into one tolist() call, batching cluster_count+is_free into one tolist() call, and maintaining a CPU mirror of the indices tensor.

## Test plan
- [ ] Conformer RMSD tests (prealigned and Kabsch paths)
- [ ] TFD tests with >1415 conformers and ring torsions with 0 quartets
- [ ] Fused Butina produces identical cluster assignments
- [ ] BFGS converges correctly for negative-energy geometries
- [ ] Similarity tests on sm_80+ and pre-Ampere hardware
- [ ] Eigensolver convergence unchanged across multiple calls on same instance

### DIFF
--- a/nvmolkit/clustering.py
+++ b/nvmolkit/clustering.py
@@ -161,10 +161,9 @@ def fused_butina(
 
             # Batch max and last-argmax into one D2H transfer (sync 1 of 2).
             neigh_flipped = neigh.flip(0)
-            batch_ma = torch.stack([
-                neigh.max().to(torch.int64),
-                neigh_flipped.argmax().to(torch.int64),
-            ]).tolist()
+            batch_ma = torch.stack(
+                [neigh.max().to(torch.int64), neigh_flipped.argmax().to(torch.int64)]
+            ).tolist()
             max_val = int(batch_ma[0])
             if max_val == 0:
                 break
--- a/nvmolkit/clustering.py
+++ b/nvmolkit/clustering.py
@@ -182,11 +182,13 @@ def fused_butina(
 
             cluster_sizes.append(cc[0])
 
-            # Update GPU tensors using CPU-derived mask (H2D, cheap).
-            is_free_tensor = torch.tensor(is_free_host, dtype=torch.bool, device=device)
-            x, y = x[is_free_tensor, :].contiguous(), x[~is_free_tensor, :].contiguous()
-            indices = indices[is_free_tensor].contiguous()
-            neigh = neigh[is_free_tensor].contiguous()
+            # is_free is already updated in-place on GPU by extract_cluster_and_singletons;
+            # use it directly to avoid a H2D→CPU→H2D roundtrip.
+            # is_free_host (from the combined download above) is still needed for indices_host.
+            y = x[~is_free, :].contiguous()
+            x = x[is_free, :].contiguous()
+            indices = indices[is_free].contiguous()
+            neigh = neigh[is_free].contiguous()
             is_free = torch.ones(x.shape[0], dtype=torch.bool, device=x.device)
             indices_host = [idx for idx, keep in zip(indices_host, is_free_host) if keep]
 
--- a/nvmolkit/clustering.py
+++ b/nvmolkit/clustering.py
@@ -161,9 +161,7 @@ def fused_butina(
 
             # Batch max and last-argmax into one D2H transfer (sync 1 of 2).
             neigh_flipped = neigh.flip(0)
-            batch_ma = torch.stack(
-                [neigh.max().to(torch.int64), neigh_flipped.argmax().to(torch.int64)]
-            ).tolist()
+            batch_ma = torch.stack([neigh.max().to(torch.int64), neigh_flipped.argmax().to(torch.int64)]).tolist()
             max_val = int(batch_ma[0])
             if max_val == 0:
                 break
--- a/nvmolkit/clustering.py
+++ b/nvmolkit/clustering.py
@@ -139,6 +139,8 @@ def fused_butina(
         n_start = x.shape[0]
         device = x.device
         indices = torch.arange(n_start, dtype=torch.int32, device=device)
+        # CPU mirror of indices avoids a D2H sync to record each centroid.
+        indices_host = list(range(n_start))
         cluster_count = torch.zeros(2, dtype=torch.int32, device=device)
         cluster_count[1] = n_start - 1
         cluster_indices = torch.zeros(n_start, dtype=torch.int32, device=device)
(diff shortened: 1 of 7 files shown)


src/minimizer/bfgs_minimize.cu
```diff
@@ -908,7 +908,9 @@ __global__ void updateDGradKernel(const double  gradTol,
   double blockMax = cub::BlockReduce<double, 128>(tempStorage).Reduce(localMax, cubMax());
 
   if (idxWithinSystem == 0) {
-    const double term = max(energies[sysIdx] * gradScales[sysIdx], 1.0);
+    // Use |energy| so negative-energy geometries (possible mid-minimization)
+    // don't clamp the denominator to 1 and artificially tighten convergence.
+    const double term = max(fabs(energies[sysIdx]) * gradScales[sysIdx], 1.0);
```
scal444: I think you've uncovered a real bug in RDKit which we reproduce [here](https://github.com/rdkit/rdkit/blob/master/Code/Numerics/Optimizer/BFGSOpt.h#L255).

We don't have an official policy on this, but typically when there's a non-massive RDKit bug, we prefer to match the rdkit behavior, file a bug upstream, and then conditionally change our behavior based on which RDKit we're compiling against.

So for this, I'd suggest filing an issue on the rdkit forum (and then you or I can put up a PR). Then change this line to a TODO to take the abs value when rdkit fixes upstream, which we'll gate at compile time against RDKit version.

nvmolkit/clustering.py
```diff
@@ -149,33 +151,53 @@ def fused_butina(
...
+
+            # Batch cluster_count and is_free into one D2H transfer (sync 2 of 2).
+            # combined[:2] = cluster_count, combined[2:] = is_free as int32.
+            combined = torch.cat([cluster_count, is_free.to(torch.int32)]).tolist()
+            cc = [int(combined[0]), int(combined[1])]
+            is_free_host = [bool(v) for v in combined[2:]]
+
+            cluster_sizes.append(cc[0])
+
+            # Update GPU tensors using CPU-derived mask (H2D, cheap).
```
scal444: There's no need to go back from is_free_host to is_free_tensor here, you can just use the existing is_free on device without a host_trip

```python
x, y = x[is_free, :].contiguous(), x[~is_free, :].contiguous()
indices = indices[is_free].contiguous()
neigh = neigh[is_free].contiguous()
```

nvmolkit/clustering.py
```diff
@@ -149,33 +151,53 @@ def fused_butina(
...
                 x, id_max, is_free, neigh, cluster_count, cluster_indices, threshold, indices, metric=metric
             )
-            cluster_sizes.append(cluster_count[0].item())
-            x, y = x[is_free, :].contiguous(), x[~is_free, :].contiguous()
-            indices = indices[is_free].contiguous()
-            neigh = neigh[is_free].contiguous()
+
+            # Batch cluster_count and is_free into one D2H transfer (sync 2 of 2).
+            # combined[:2] = cluster_count, combined[2:] = is_free as int32.
+            combined = torch.cat([cluster_count, is_free.to(torch.int32)]).tolist()
```
scal444, line 177: Rather than combining to a new tensor each time, could you pull an intermediate variable out of the loop and populate the fields?

src/conformer_rmsd.cu
```diff
@@ -99,15 +99,33 @@ __device__ __forceinline__ double det3x3(const double* H) {
 // ---------------------------------------------------------------------------
 
 constexpr int kRmsdBlockSize = 128;
-using RmsdBlockReduceT       = cub::BlockReduce<double, kRmsdBlockSize>;
+constexpr int kRmsdWarps     = kRmsdBlockSize / 32;  // 4 warps per block
+
+// Warp-level double-precision sum via shuffle; lane 0 holds the result.
+__device__ __forceinline__ double warpSumDouble(double v) {
```
scal444: Rather than rolling it ourselves, prefer to use cub::WarpReduce should will do the same thing under the hood.

src/similarity_kernels.cu
```diff
@@ -65,6 +65,29 @@ bool supportsTensorOps(const int major, const int minor) {
...
+  if (device >= 0 && device < kMaxDevices && g_tensorOpsCache[device] != 0) {
+    return g_tensorOpsCache[device] == 2;
+  }
+  cudaDeviceProp props;
+  cudaGetDeviceProperties(&props, device);
+  const bool result = supportsTensorOps(props.major, props.minor);
+  if (device >= 0 && device < kMaxDevices) {
+    g_tensorOpsCache[device] = result ? 2 : 1;
+  }
+  return result;
```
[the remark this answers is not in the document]
scal444: Agree with the greptile comment here

scal444:
Thanks for the contribution!

The TFD, triangle smooth, eigensolver, and similarity fixes are good to go except for the Greptile comment. I measured a modest speedup in our similarity benchmarks. 

The Butina change is good, and showed major (nearly 2x) speedups in most of our benchmarks. I identified one regression which when fixed adds another 5-10% speedup, and made a few suggestions.

The Conformer RMSD change also works, with more modest (~10%) speedups. I think it would be worthwhile to experiment with a fused reduction, creating a combined datatype with all the concurrent reductions and doing a block reduce on that, but this is a real improvement so can go in without that experiment.

scal444 (changes requested):

src/minimizer/bfgs_minimize.cu
mooreneural: Sounds logical, we can update this later when they do.

Here is my PR for RDKit https://github.com/rdkit/rdkit/pull/9298

nvmolkit/clustering.py
mooreneural: Eliminated the is_free_tensor H2D→CPU→H2D roundtrip. is_free is already updated in-place on GPU by extract_cluster_and_singletons, so we now index x, indices, and neigh directly from it. is_free_host (still downloaded via the combined transfer) is only used for the CPU indices_host mirror update.

src/conformer_rmsd.cu
mooreneural: Swapped the hand-rolled warpSumDouble (__shfl_down_sync loop) for cub::WarpReduce<double>. Behaviour and generated PTX are identical since CUB uses shuffles internally on full 32-thread warps, it's purely a CUB-idiom consistency fix as you suggested.

src/minimizer/bfgs_minimize.cu
mooreneural: Reverted the fabs() on the energy denominator back to RDKit's signed behaviour. Left a // TODO: file upstream RDKit bug comment so we don't lose track of it, and noted it should be version-gated when/if RDKit merges a fix (matching the pattern already in scaleGradKernel).

src/minimizer/bfgs_minimize.cu
```diff
@@ -908,6 +908,10 @@ __global__ void updateDGradKernel(const double  gradTol,
   double blockMax = cub::BlockReduce<double, 128>(tempStorage).Reduce(localMax, cubMax());
 
   if (idxWithinSystem == 0) {
+    // Matches RDKit's signed-energy convergence denominator in ForceField::minimize.
+    // Negative-energy geometries can clamp this to 1, artificially tightening the
+    // check — but fixing it unconditionally diverges from reference behavior.
+    // TODO: file upstream RDKit bug; gate fix on RDKit version once merged there.
     const double term = max(energies[sysIdx] * gradScales[sysIdx], 1.0);
```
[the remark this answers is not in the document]
mooreneural: Updated in the beginning of the PR: bfgs_minimize.cu Convergence denominator kept as RDKit's signed-energy form to preserve reference behavior. Added TODO to file upstream RDKit bug and gate the fix on RDKit version once merged (matching the pattern already used in scaleGradKernel).

scal444 (approved):

scal444:
This looks great, just needs one more ruff format.

PR #179, opened 2026-05-26
Improve autotune batch size and CPU count scanning

scal444:
Autotune now steps in 64 element increments by default, cutting down on the search space. CPU space is now physical core limited by default.

### DIFF
--- a/nvmolkit/autotune/_ff_common.py
+++ b/nvmolkit/autotune/_ff_common.py
@@ -24,6 +24,7 @@ not consumed by the C++ batched-FF code path (threads come from
 from __future__ import annotations
 
 import os
+import re
 from typing import Iterable, Optional
 
 from rdkit.Chem import Mol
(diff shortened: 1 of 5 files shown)


nvmolkit/autotune/_ff_common.py
```diff
@@ -49,17 +50,37 @@ def coerce_gpu_ids(gpuIds: Optional[Iterable[int]]) -> list[int]:
 
 
 def cpu_count() -> int:
-    """Return the usable CPU core count, with a floor of 1."""
+    """Return the usable physical CPU core count, with a floor of 1.
+
+    Falls back to ``os.cpu_count()`` (logical) if ``/proc/cpuinfo`` is missing.
```
evasnow1992, line 55: Not a blocker, but if users run this inside Docker with `--cpus=N` or cgroup CPU quotas, both `os.cpu_count()` and `/proc/cpuinfo` will reflect the host hardware rather than the container’s effective CPU allocation.

evasnow1992 (approved):
LGTM. Only one minor comment. Thanks!

nvmolkit/autotune/_ff_common.py
scal444: We can live with that, user can also set bounds explicitly.

PR #180, opened 2026-05-26
Add early-exit option for RDKit benchmarks that operates on a specific time budget

scal444:
Calculates throughput after running for up to X amount of time.

Refactor of substruct rdkit code to reduce churn, but it's still unfortunately verbose due to indentation issues.

### DIFF
--- a/benchmarks/substruct_bench.py
+++ b/benchmarks/substruct_bench.py
@@ -930,9 +930,11 @@ def main():
             nvmolkit_config_source = config_source if name == "nvmolkit" else "N/A"
             if is_rdkit:
                 rdkit_match_mode, rdkit_threads = rdkit_variant_meta[name]
+                rdkit_max_seconds = args.rdkit_max_seconds
             else:
                 rdkit_match_mode = "N/A"
                 rdkit_threads = "N/A"
+                rdkit_max_seconds = "N/A"
             throughput = (pairs_done * 1000.0 / avg_ms) if avg_ms > 0 else 0.0
             vs_rdkit = (throughput / baseline_throughput) if (not is_rdkit and baseline_throughput > 0) else "N/A"
             csv_rows.append(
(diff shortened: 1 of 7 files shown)


benchmarks/substruct_bench.py
```diff
@@ -848,6 +955,10 @@ def main():
                     rdkit_match_mode,
                     avg_ms,
                     std_ms,
+                    pairs_done,
+                    args.rdkit_max_seconds,
+                    throughput,
```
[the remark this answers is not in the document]
scal444: @greptileai, this is fixed, please recheck

benchmarks/substruct_bench.py
```diff
@@ -268,57 +292,89 @@ def bench_rdkit_substruct(
...
-                    mol_results = []
-                    for query in queries:
-                        matches = mol.GetSubstructMatches(query, params)
-                        mol_results.append(matches)
-                    results_data.append(mol_results)
-
-    avg_ms, std_ms = time_it(run, runs)
-    return avg_ms, std_ms, results_data
+            pairs_done_this_run = 0
+            run_deadline = Deadline(max_seconds)
```
evasnow1992: I may be misunderstanding the logic here, but is this `Deadline` initialization redundant? It looks like `time_it_bounded`called below also creates and manages its own `Deadline` object.

benchmarks/substruct_bench.py
```diff
@@ -327,34 +383,18 @@ def run():
             lib.AddMol(mol)
 
         results_data = [[None] * num_queries for _ in range(num_mols)]
+        pairs_done_this_run = 0
+        run_deadline = Deadline(max_seconds)
```
evasnow1992: Same here. Wonder if this is redundant.

benchmarks/substruct_bench.py
scal444: They were both relevant, one checks the in-run time, one decides whether to start a new run. 

But we really didn't need both, the same object could be queried multiple places. Fixed.

evasnow1992 (approved):

PR #181, opened 2026-05-27
Reduce conformer bench preprocessing times via single embed + jitter

scal444:
Instead of running ETKDG on all conformers, embed one conformer, duplicate, jitter coordinates. Improves bottleneck for TFD, conformer RMSD, and ff minimization on the fly generation

### DIFF
--- a/benchmarks/tfd_bench.py
+++ b/benchmarks/tfd_bench.py
@@ -32,6 +32,7 @@ Example:
 """
 
 import argparse
+import multiprocessing
 import os
 import pickle
 import sys
@@ -40,8 +41,9 @@ from typing import List, Tuple
 
 import pandas as pd
 import torch
+from bench_utils import embed_and_jitter, load_smiles
 from rdkit import Chem
-from rdkit.Chem import AllChem, TorsionFingerprints
+from rdkit.Chem import TorsionFingerprints
 
 import nvmolkit.tfd as nvmol_tfd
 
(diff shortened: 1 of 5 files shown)


benchmarks/tfd_bench.py
```diff
@@ -72,30 +74,31 @@ def time_it(func, runs: int = 3, warmups: int = 1) -> Tuple[float, float]:
...
-
-    if len(conf_ids) < 2:
-        return None
-
-    mol = Chem.RemoveHs(mol)
-    return mol
+    if num_confs < 2:
+        raise ValueError(f"num_confs must be >= 2 for TFD, got {num_confs}")
+    workers = num_workers if num_workers > 0 else max(1, multiprocessing.cpu_count() // 2)
+    return embed_and_jitter(
```
evasnow1992, line 93: I’m a little concerned about replacing `EmbedMultipleConfs` with `embed_and_jitter`. If I understand correctly, the latter may generate conformers with near-identical torsion angles and without meaningful dihedral rotations. In that case, the TFD values could collapse close to 0, where the numerical comparison between RDKit and nvMolKit TFD implementations may become more variable or less informative.

evasnow1992 (approved):
One comment, but no code validity issues. Changes look good to me.

benchmarks/tfd_bench.py
scal444: Valid concern - I would expect that the values are different enough but I'll confirm with a quick experiment.

scal444: I added an additional spread parameter to avoid RMSD clumps, now both TFD and RMSD have reasonable spreads compared to ETKDG
image

PR #183, opened 2026-05-28
Gate BFGS gradient-convergence denominator fix on RDKit >= 2026.03

mooreneural:
rdkit/rdkit#9298 (merged into RDKit 2026.03) fixed a bug where a negative energy value caused max(energy * gradScale, 1.0) to clamp to 1, artificially tightening the gradient-tolerance convergence test mid-minimisation. Force fields with stabilising electrostatic or dispersion terms (MMFF94, UFF) can produce negative intermediate energies, so this affected real workloads.

Apply the same fix in both nvMolKit BFGS paths, gated on the linked RDKit version so that behaviour is unchanged when built against older RDKit:

- src/minimizer/bfgs_minimize.cu       (updateDGradKernel, batched path)
- src/minimizer/bfgs_minimize_permol_kernels.cu (updateDGrad, per-mol path)

Mirrors the version-conditional pattern already used for kRdkitHasGradScaleFix in scaleGradKernel.

### DIFF
--- a/src/minimizer/bfgs_minimize.cu
+++ b/src/minimizer/bfgs_minimize.cu
@@ -908,11 +908,13 @@ __global__ void updateDGradKernel(const double  gradTol,
   double blockMax = cub::BlockReduce<double, 128>(tempStorage).Reduce(localMax, cubMax());
 
   if (idxWithinSystem == 0) {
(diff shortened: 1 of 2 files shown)


scal444 (approved):

scal444:
Awesome, thank you!

PR #185, opened 2026-05-28
Update most links for domain name change

scal444:
Exceptions are:

1. Changelog - those links redirect fine
2. Docker container instructions, we need to actually push to the new package registry, will do offline then follow up with a docs change


evasnow1992 (approved):
Thank you for updating the paths. LGTM.

PR #187, opened 2026-06-01
Use throughput data for Conformer RMSD benchmark

scal444:
Before, it had been based on single smiles. Modified to be more like our other benchmarks, and added RDKit early exit. The early exit is a bit annoying right now because we can't use `time_it`, added a tracking bug

### DIFF
--- a/benchmarks/conformer_rmsd_bench.py
+++ b/benchmarks/conformer_rmsd_bench.py
@@ -23,6 +23,7 @@ CPU GetConformerRMSMatrix across varying conformer counts.
 import argparse
 import csv
 import multiprocessing as mp
+import statistics
 import time
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


benchmarks/conformer_rmsd_bench.py
```diff
@@ -13,163 +13,285 @@
...
+            cap_label = f"cap={rdkit_max_seconds:.0f}s" if rdkit_max_seconds > 0 else "no cap"
+            print(f"  RDKit CPU (single-threaded, {cap_label}):")
+            # TODO: replace this hand-rolled warmup/sample/median loop with time_it once
+            # time_it can consume a Deadline and truncate a run mid-workload.
+            # https://github.com/NVIDIA-BioNeMo/nvMolKit/issues/186
+            bench_rdkit_batch(payloads, rdkit_max_seconds)  # warmup
+            samples = [bench_rdkit_batch(payloads, rdkit_max_seconds) for _ in range(3)]
+            samples.sort(key=lambda pair: pair[0] / max(pair[1], 1))
+            rdkit_time_s, rdkit_done = samples[len(samples) // 2]
+            rdkit_std_s = statistics.stdev([elapsed for elapsed, _ in samples]) if len(samples) > 1 else 0.0
```
evasnow1992: Looks like there may be an inconsistency between the median and std calculations. Since `pair[0]` is normalized by `pair[1]` when selecting the median (`rdkit_time_s`), that value appears to represent per-molecule time. However, `rdkit_std_s` is still computed directly from the raw elapsed times?

scal444: Fixed

evasnow1992 (approved):

PR #188, opened 2026-06-01
Butina clustering bench improvements

scal444:
* Fixed an error when tanimoto similarity of 1.1 was an option, it causes an error in some paths. 1.0 still does the same thing without the error.

* Standardizes measurements for different modes. When we run the fused or low mem rdkit/nvmolkit versions, the proper comparison is against the tanimoto matrix THEN Butina for the non-fused versions. However, when directly comparing non-fused, it makes more sense to leave out the tanimoto aspect.


evasnow1992 (approved):

PR #189, opened 2026-06-01
Add more skip nvMolKit flags to benchmarks

scal444:
The similarity one is annoying because it uses pyperf, which actually subprocesses the entire file for multiple runs.


evasnow1992 (approved):
Thank you for adding these flags. The changes look good to me.
We may consider consolidating similar flag names across the benchmark scripts for consistency as a future follow-up.

PR #190, opened 2026-06-01
Use multiprocessing in ETKDG benchmark validation

scal444:
It was an unneeded bottleneck.


evasnow1992 (approved):
Since --rdkit_threads is now overloaded, we may want to update its help text on line 274 to reflect the expanded behavior. Other changes look good to me.

PR #191, opened 2026-06-01
Major Cmake / C++ include refactor

scal444:
Removed all of our manual in-tree include CMake code. All includes are now relative to the project root, with most of them under `src/`. 

Added a new formatter checker to CI to avoid backslide.


admin/include_check.py
```diff
@@ -0,0 +1,315 @@
...
+        return line, None
+    include_str = after[:end]
+    trailing = after[end + 1:]
+
+    rewritten, reason = resolve_include(
+        project_root, file_path, include_str, header_index
+    )
+    if rewritten is None:
+        # Already-correct quoted form, system include, or third-party.
+        if close_char == ">" and rewritten is None and reason is None:
```
evasnow1992: `rewritten is None` this line seems to be redundant given it's inside the if block from line 189

evasnow1992 (approved):
One minor comment. Other changes look good to me. thanks!

admin/include_check.py
scal444: Done

Issue #197, opened 2026-06-06
Rare Morgan Fingerprint off-by-1-bit error


scal444:
A cache buffer that needed to be zeroed between batch iterations is  not, leading to rare (~1 in 10k) cases of subsequent batches with fingerprints off by (usually) 1 bit

scal444:
Found by @vfscalfani thanks to #195

scal444:
Fixed, will be in v0.5.1

PR #198, opened 2026-06-06
Fix gh 197

scal444:
Zero out a buffer that needed to be zero'd out every thread iteration, not just at the beginning of the batch. Wrote a reliable reproducer.


scal444:
Fixes #197

evasnow1992 (approved):
Thank you for fixing this! Changes look good to me.

PR #204, opened 2026-06-11
Fix GH issue 203  - uncaught exception in OpenMP

### LINKED ISSUE #203: hasSubstuctMatch kernel crash with disconnected SMARTS
Hi, 

When trying the following minimal example:

```
smarts = "C.C"
smiles = ['CC', 'CCC', 'CCCC']

mols = [Chem.MolFromSmiles(smi) for smi in smiles]
pattern = Chem.MolFromSmarts(smarts)

config = SubstructSearchConfig(
    batchSize=1024,
    workerThreads=-1,
    preprocessingThreads=1,
    maxMatches=1,
    uniquify=False,
)

result_test = hasSubstructMatch(
    targets=mols,
    queries=[pattern],
    config=config,
)

print(f"GPU count: {result_test[:, 0].sum()}")

```

I get a helpful and expected output:

```
RuntimeError: Fragment queries (disconnected SMARTS patterns) are not supported. Query has 2 disconnected components: C.C
```

However, if I change the preprocessingThreads to >1 or -1, I get a kernel crash (using Jupyter Notebook), instead of the RunTimeError message.

Thanks, 

Vin

scal444:
We needed to catch and rethrow outside the openmp loop. Fixes #203


evasnow1992 (approved):
Thank you for fixing this. The current changes look good to me. We may, however, have similar issues elsewhere in `molecules.cpp`, such as around lines 2078–2086. It might be worth doing a quick scan with follow-up PRs if any related cases are found.

Issue #202, opened 2026-06-10
nvmolkit.embedMolecules does not generate a conformer for every molecule in the set


Yannlefrancois:
Hi,
i'm using nvmolkit to embeded conformation in a large amount of molecule, i'm trying my current code on a small subset of molecule and i see that it only embeded comformation in part of the molecules.

here is my code 

import nvmolkit.embedMolecules as embed
from rdkit import Chem
from rdkit.Chem import Descriptors
from rdkit.Chem import AllChem

        suppl=Chem.MultithreadedSmilesMolSupplier(f"{arg.temp_file_dir}/{file}", "\t", 0, 1, True, True,numWriterThreads=48)
        
        
        # create the list in advance to avoid unecessary memory requirement upon list expension using append

        mol_list=[None for i in range(arg.batch_size)]
        mol_id=[None for i in range(arg.batch_size)]
        mol_counter=-1
        for mol_1 in suppl:
            mol_counter+=1
            if mol_1 is not None:
                try:
                    mol_list[mol_counter]=Chem.rdmolops.AddHs(mol_1)
                    mol_id[mol_counter]=mol_1.GetProp(arg.property_name)
                except :
                    pass
        

  
            params = Chem.rdDistGeom.ETKDG()
            params.useRandomCoords = True
            hardware_opts = HardwareOptions(
            preprocessingThreads=48,
            batchSize=-1,
            batchesPerGpu=-1,
            gpuIds=[0, 1])
            embed.EmbedMolecules(mol_list, params, confsPerMolecule=1, hardwareOptions=hardware_opts)

and if i check with mol.GetNumConformers() a non negligable amount as 0 500 out of 100000

is there anything that could explain that ? 

rdkit version = 2026.03.1

nvmolkit 0.5.0

best 
yann

scal444:
Hi Yann, thanks for the report.

Is it consistently the same molecules? Does RDKit also fail on these molecules?

In general both RDKit and nvMolKit can definitely fail on poorly formed molecules, molecules where AddHs failed, or occasionally on good molecules that are hard to generate. Part of what controls success rate for the latter case is the max attempts param and the number of conformers you are trying to generate. 

If you can provide me reproducing molecules or point me to where you're sourcing them from, I can follow up to see if it's a real bug!

Yannlefrancois:
[test1.txt](https://github.com/user-attachments/files/28801126/test1.txt)
[...]

PR #205, opened 2026-06-11
Fix dead RDKIT_NEW_FLAG_API gate in ETKDG chirality and bounds code

scal444:
GH issue 202: a non-negligible fraction of molecules embedded to zero conformers. RDKIT_NEW_FLAG_API was a constexpr bool but consumed with #if, which the preprocessor evaluates as 0, and etkdg_stages and rdkit_dist_geom_flattened never linked nvmolkit_versions so versions.h was not even visible. The gate was therefore always off.

findChiralSets never set d_structureFlags and loadChiralDataset never read it, so fused-small-ring tetrahedral centers used volScale 1.0 instead of the 0.25 RDKit applies and were deterministically rejected at the tetrahedral check. The same gate guards the constrained tight-bounds branch in addLongRangeDistanceConstraints, which assigns to l and u; those are now non-const so the branch compiles.


scal444:
Fixes #202

evasnow1992 (approved):

Issue #201, opened 2026-06-08
Investigate poor LLM usage of conformer RMSD


scal444:
Internal testing indicates models don't often correctly use the conformer RMSD API, even when using our skill. This may indicate inconsistency in the API docs or other documentation.

PR #206, opened 2026-06-11
Improve conformer RMSD workflow ergonomics

scal444:
Fixes an incorrect documentation note that hints the output format is a square matrix. Actually now provides the path to clustering after conf RMSD.


scal444:
This goes after #201

evasnow1992 (approved):
Changes look good to me.

Issue #200, opened 2026-06-08
Better compatibility for array APIs


scal444:
Butina and fused Butina should all take asyncGpuResult, torch tensor, numpy array. We should scan the rest of our APIs too, but right now we're inconsistent at least with these.

PR #207, opened 2026-06-11
Unify array inputs for several functions

scal444:
Allow any of asyncGpuResult, CPU tensor, GPU tensor, and numpy array, with conversions as necessary. Use proper torch stream interactions


scal444:
This fixes #200

nvmolkit/clustering.py
```diff
@@ -80,24 +92,25 @@ def butina(
...
+    with torch.cuda.stream(active_stream):
+        distance_matrix_tensor = _as_cuda_tensor("distance_matrix", distance_matrix, stream=active_stream)
+        distance_matrix_tensor = _check_distance_matrix("distance_matrix", distance_matrix_tensor)
+        result = _clustering.butina(
+            distance_matrix_tensor.__cuda_array_interface__,
+            cutoff,
+            neighborlist_max_size,
+            return_centroids,
+            active_stream.cuda_stream,
+        )
```
evasnow1992, line 105: Should we also add result._input_refs = (distance_matrix_tensor) like in the similarity implementation below to keep the device buffer alive after return?

evasnow1992 (approved):
Changes look good to me. One minor comment, not a blocker.

nvmolkit/clustering.py
scal444: Good thought, but that one was needed because the similarity API is completely async and the the GC might collect if no references are there before the function is done. Butina returns GPU data if asked for but syncs internally, so the function will return while the reference is still active.

PR #212, opened 2026-06-15
Fix MMFF/UFF optimization modules failing to import standalone

### LINKED ISSUE #211: Brittle import order dependency for hardware options
We need to untangle pybind deps between hardware options (in types.py) and our minimizers, sometimes importing in one order leads to an error.

scal444:
The _mmffOptimization and _uffOptimization bindings declared BatchHardwareOptions default argument values. Boost.Python converts default values to Python when a function is registered, which needs a to-Python converter that only _embedMolecules registers, so importing either module before nvmolkit.types raised "No to_python converter found for nvMolKit::BatchHardwareOptions". The Python wrappers already supply every argument, so the native defaults are dropped, and a regression test imports each module first in a fresh interpreter.

Fixes #211

### DIFF
--- a/nvmolkit/_mmffOptimization.pyi
+++ b/nvmolkit/_mmffOptimization.pyi
@@ -3,7 +3,7 @@ from rdkit.Chem import Mol
 
 def MMFFOptimizeMoleculesConfs(
     molecules: List[Mol],
-    maxIters: int = 200,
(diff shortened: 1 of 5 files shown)


evasnow1992 (approved):
Changes look good to me.

PR #217, opened 2026-06-29
Fix DG torsion sixth-order gradient, adjust tests for RDKit mismatch

scal444:
The FIRE minimizer is more sensitive to incorrect gradient terms than BFGS. In advance of FIRE, making this fix, and adjusting our tests to separate out the incorrect rdkit term.


scal444:
PR 1 out of ~ 8 for FIRE support, #216

evasnow1992 (approved):
Thank you for fixing this known issue and detaching the related tests from RDKit's reference values. Changes look good to me. Thanks!

PR #218, opened 2026-06-29
[Draft] Full FIRE implementation

scal444:


scal444:
This is for reference, each commit will be staged separately for review for #216

PR #219, opened 2026-06-30
[FIRE 2/N] FIRE kernels and minimizer

scal444:
Largest PR for #216. Implements the FIRE kernels and minimizer, based on ASE implementation with a few extra stop criteria options.


evasnow1992 (approved):
Thank you for composing the FIRE kernels and minimizer. Changes look good to me.

PR #220, opened 2026-07-03
 Adds support for `reordering=False` in Butina clustering

### LINKED ISSUE #192: Non-reordering Butina clustering
We only support the `reordering=True` variant, which is the more rigorous and expensive version. It should be easy to implement a version where initial neighbor counts are frozen.

Matthew-Neba:
## Summary

  Adds support for `reordering=False` in Butina clustering.  Fixes #192 

  ## Changes

  - Added a `reordering` option to the Python and C++ Butina APIs.
  - Implemented fixed-order Butina clustering for `reordering=False`.
  - Added tests and ensured they passed
  - Updated the benchmark to optionally run nvMolKit Butina with reordering enabled, disabled, or both.

### DIFF
--- a/nvmolkit/clustering.cpp
+++ b/nvmolkit/clustering.cpp
@@ -43,7 +43,7 @@ BOOST_PYTHON_MODULE(_clustering) {
       if (!streamOpt) {
         throw std::invalid_argument("Invalid CUDA stream");
       }
-      auto                             stream                = *streamOpt;
+      auto                             stream  = *streamOpt;
       // Extract boost::python::tuple from dict['shape']
       boost::python::tuple             shape   = boost::python::extract<boost::python::tuple>(distanceMatrix["shape"]);
       const size_t                     matDim1 = boost::python::extract<size_t>(shape[0]);
@@ -58,13 +58,8 @@ BOOST_PYTHON_MODULE(_clustering) {
         centroids.setStream(stream);
       }
       const auto centroidSpan = returnCentroids ? toSpan(centroids) : cuda::std::span<int>{};
-      const int numClusters = nvMolKit::butinaGpu(matSpan,
-                                                  toSpan(clusterIds),
-                                                  cutoff,
-                                                  neighborlistMaxSize,
-                                                  centroidSpan,
-                                                  stream,
-                                                  reordering);
+      const int numClusters =
+        nvMolKit::butinaGpu(matSpan, toSpan(clusterIds), cutoff, neighborlistMaxSize, centroidSpan, stream, reordering);
       if (returnCentroids) {
         auto clusterArray  = nvMolKit::makePyArray(clusterIds, boost::python::make_tuple(matDim1));
         auto centroidArray = nvMolKit::makePyArray(centroids, boost::python::make_tuple(numClusters));
--- a/nvmolkit/clustering.cpp
+++ b/nvmolkit/clustering.cpp
@@ -37,8 +37,8 @@ BOOST_PYTHON_MODULE(_clustering) {
         const double               cutoff,
         const int                  neighborlistMaxSize,
         const bool                 returnCentroids,
-        std::uintptr_t             streamPtr,
-        const bool                 reordering) -> boost::python::object {
+        const bool                 reordering,
+        std::uintptr_t             streamPtr) -> boost::python::object {
       auto streamOpt = nvMolKit::acquireExternalStream(streamPtr);
       if (!streamOpt) {
         throw std::invalid_argument("Invalid CUDA stream");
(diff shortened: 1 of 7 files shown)


src/butina.cu
```diff
@@ -1050,24 +1323,56 @@ __global__ void thresholdDistanceMatrixKernel(const double* __restrict__ matrix,
...
 }  // namespace
 
 [[maybe_unused]] int butinaGpu(const cuda::std::span<const double> distanceMatrix,
                                const cuda::std::span<int>          clusters,
                                const double                        cutoff,
                                const int                           neighborlistMaxSize,
                                const cuda::std::span<int>          centroids,
-                               cudaStream_t                        stream) {
+                               cudaStream_t                        stream,
+                               const bool                          reordering) {
```
scal444: Minor - prefer to have the stream be the last optional parameter, we try to have `<mandatory>, <optional>, <gpu>` args in order, though I'm sure we're not 100% consistent with that.

nvmolkit/clustering.cpp
```diff
@@ -37,7 +37,8 @@ BOOST_PYTHON_MODULE(_clustering) {
         const double               cutoff,
         const int                  neighborlistMaxSize,
         const bool                 returnCentroids,
-        std::uintptr_t             streamPtr) -> boost::python::object {
+        std::uintptr_t             streamPtr,
+        const bool                 reordering) -> boost::python::object {
```
scal444: Same here with the stream param ordering.

nvmolkit/clustering.py
```diff
@@ -70,13 +71,16 @@ def butina(
...
+                distance is less than or equal to this cutoff.
         neighborlist_max_size: Maximum size of the neighborlist used for small cluster
                               optimization. Must be 8, 16, 24, 32, 64, or 128. Larger values
                               allow parallel processing of larger clusters but use more
                               shared memory.
         return_centroids: Whether to return centroid indices for each cluster.
         stream: CUDA stream to use. If None, uses the current stream.
+        reordering: Whether to update neighbor counts among unassigned items
+                    after each cluster is formed. The default matches the
+                    existing nvMolKit behavior and RDKit's ``reordering=True``.
```
scal444: You're right to not silently change our default and to call out the delta from rdkit's default, but instead of referring to nvMolKit's previous state, how about we just say "Note: Defaults to True, while RDKit's Butina.ClusterData defaults to False.

src/butina.cu
```diff
@@ -447,6 +447,140 @@ void renumberClustersBySize(const cuda::std::span<int> clusters,
...
+  const AsyncDeviceVector<uint8_t> sortTemp(sortTempBytes, stream);
+
+  // 3) run the radix sort
+  cub::DeviceRadixSort::SortKeysDescending(
+    sortTemp.data(), sortTempBytes, sortKeys.data(), sortedKeys.data(), numPoints, 0, sizeof(uint64_t) * 8, stream);
+  cudaCheckError(cudaGetLastError());
+}
+
+// Select the next unassigned candidate for parallel row assignment. Note that this task is not inherently parallelizable. We walk the vector element by element until we find a valid centroid. If done in parallel, we could accidentallly select the wrong elements as centroids.
+__global__ void prepareFixedOrderCandidateKernel(const cuda::std::span<const uint64_t> sortedKeys,
```
scal444, line 511: I think this could be parallelized - We want the first nonzero element each time, right? Or the first valid one if it's a different flag, though we could consolidate it on this path. So, starting from cursor, each thread in a block could check (cursor + 0, cursor + 1, cursor +2), and if empty, check the next section. 

I looked around for a good way to do this. I don't see any CUDA equivalent of std::find_first_of, but you could do it via a blockReduce where you set the value == the index if there's a match, and a sentinel with no match. Something like

```cpp
    const int index = threadIdx.x; 
    const int candidate = (index < count && data[index + cursorIdx] != 0) ? index : INT_MAX;
    const int firstIndex = BlockReduce(tempStorage).Reduce(candidate, cub::Min());
    // check here if it's INT_MAX to see if something was found in that range. 
```

scal444:
Thanks for the contribution! This looks great, and we can definitely get it in for v0.6. A few comments, mostly minor.

Matthew-Neba:
Thanks for all the help! I have fixed the issues you pointed out. I have managed to parallelize that kernel and fixed the other minor issues aswell. I could indeed use the blockReduce idea you mentioned and it works well.

src/butina.cu
Matthew-Neba: Rewrote the kernel, including parallelization now.

scal444:
Looks great! Can you install ruff 0.15.8 and clang-format-17, and then run admin/run_clang_format.sh and admin/run_ruff_format.sh to get it to pass the linters?

scal444:
Or i can do it and push a commit if you're ok with that

Matthew-Neba:
No worries, I ran the formatters. I also fixed an issue that was leading to the forge build failing. Hopefully all loose ends are tied now.

scal444:
Great, thank you!

scal444 (approved):

PR #225, opened 2026-07-09
Expose FIRE minimizer Python APIs

scal444:
PR 6 for #216

### DIFF
--- a/nvmolkit/batchedForcefield.py
+++ b/nvmolkit/batchedForcefield.py
@@ -442,6 +442,7 @@ class _BatchedForcefieldBase:
             raise ValueError("minimizerKind must be 'BFGS' or 'FIRE'")
         if fireOptions is None:
             fireOptions = FireOptions()
+            fireOptions.gradTol = forceTol
         if output == CoordinateOutput.DEVICE:
             return self._native_ff.minimizeDevice(
                 maxIters,
@@ -599,7 +600,8 @@ class MMFFBatchedForcefield(_BatchedForcefieldBase):
 
         Args:
             maxIters: Maximum number of minimizer iterations.
-            forceTol: Gradient convergence tolerance.
+            forceTol: Gradient convergence tolerance. For FIRE, this value is
+                used only when ``fireOptions`` is ``None``.
             output: ``RDKIT_CONFORMERS`` (default) or ``DEVICE``.
             target_gpu: In DEVICE mode, the GPU to consolidate the result on.
                 ``None`` (the default) selects the wrapper's own GPU. The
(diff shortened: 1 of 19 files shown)


nvmolkit/uffOptimization.cpp
```diff
@@ -14,35 +14,70 @@
...
+                                                       fireOptions,
+                                                       thresholdVec,
+                                                       ignoreVec,
+                                                       hardwareOptions) :
+          nvMolKit::UFF::UFFOptimizeMoleculesConfsBfgs(molsVec, maxIters, thresholdVec, ignoreVec, hardwareOptions);
       return nvMolKit::vectorOfVectorsToList(result);
     },
     (bp::arg("molecules"),
-     bp::arg("maxIters"),
+     bp::arg("maxIters") = 1000,
```
evasnow1992: Should we avoid placing an argument with a default value before a required argument?

nvmolkit/mmffOptimization.py
```diff
@@ -179,13 +193,17 @@ def _normalize_properties(value):
...
 
-    # Call the C++ implementation
     if hardwareOptions is None:
         hardwareOptions = HardwareOptions()
     native_options = hardwareOptions._as_native()
     properties_list = _normalize_properties(properties)
     thresholds = _normalize_scalar_or_list(nonBondedThreshold, "nonBondedThreshold")
     interfrag_flags = _normalize_scalar_or_list(ignoreInterfragInteractions, "ignoreInterfragInteractions")
+    minimizer_kind = str(minimizerKind).upper()
+    if minimizer_kind not in {"BFGS", "FIRE"}:
```
evasnow1992, line 208: Should we also validate the `backend` argument at the Python level to avoid passing an invalid value to the C++ module? That way we can raise a `ValueError` instead of relying on the C++ layer to handle it.

scal444: The C++ layer already translates runtime to value error, but I think you're right that we should be consistent at what layer we handle it, fixed.

nvmolkit/uffOptimization.cpp
scal444: Done

nvmolkit/batchedForcefield.py
```diff
@@ -428,15 +429,28 @@ def _minimize(
...
-            return self._native_ff.minimizeDevice(maxIters, forceTol, -1 if target_gpu is None else int(target_gpu))
-        energies, converged = self._native_ff.minimize(maxIters, forceTol)
+            return self._native_ff.minimizeDevice(
+                maxIters,
+                forceTol,
+                -1 if target_gpu is None else int(target_gpu),
+                minimizer_kind,
+                fireOptions,
+            )
+        energies, converged = self._native_ff.minimize(maxIters, forceTol, minimizer_kind, fireOptions)
```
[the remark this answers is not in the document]
scal444: Fixed

evasnow1992 (approved):

PR #226, opened 2026-07-18
Replace fused Butina Python/Triton path with CUDA backend

Matthew-Neba:
### Summary
  - Replaced the fused Butina Python/Triton implementation with a CUDA backend.
  - Keeps the full reordering clustering loop on the GPU with CUDA Graphs
  - Avoids CPU-GPU round trips, Python loop/object overhead and materializing an `N x N` hit matrix.
  - Supports Tanimoto and cosine fingerprint similarity metrics, just like the previous version. Although, it can be extended for any fingerprint similarity metric
 - Fixed a CUDA stream synchronization bug in CPU-returning similarity operations that caused flaky results and tests.
  - Unified regular and fused Butina outputs to return device-side cluster IDs and optional centroids through a shared result type.
  - Consolidated Butina code and added shared helpers for CUDA graphs, fingerprint preparation, and Tanimoto/cosine calculations.
  - Standardized packed fingerprints on `uint32` while retaining `int32` input support 

  ### Performance
  About **12-15x faster** than the previous Python implementation on my machine.


docs/changelog.md
```diff
@@ -22,7 +22,7 @@ nvMolKit 0.5.0 adds three new GPU-accelerated APIs: Torsion Fingerprint Deviatio
 - `HardwareOptions` support for MMFF minimization, matching the ETKDG hardware-targeting API
 - Device-side output for ETKDG and forcefield optimization, allowing GPU tensors to flow between nvMolKit calls without round-tripping through host memory ([#140](https://github.com/NVIDIA-Digital-Bio/nvMolKit/issues/140))
 - Python autotuning library for the main APIs (`nvmolkit.autotune`), including ETKDG, forcefield optimization, and substructure search, with configuration serialization ([#141](https://github.com/NVIDIA-Digital-Bio/nvMolKit/issues/141))
-- Low-memory fused Butina clustering that computes Tanimoto similarities on the fly with Triton-backed kernels, avoiding the O(N²) distance matrix and enabling clustering of larger fingerprint datasets on a single GPU ([#110](https://github.com/NVIDIA-Digital-Bio/nvMolKit/issues/110))
```
scal444, line 25: We shouldn't be modifying previous versions changelogs. I'll amend our changelogs when we cut a new release.

src/fused_butina.h
```diff
@@ -0,0 +1,35 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+#ifndef NVMOLKIT_FUSED_BUTINA_H
```
scal444: Can you merge this content into the `butina.h` header and document in the header when/where to use which version?  fused_butina.cu can be added to the butina Cmake target instead of creating a new one.

It may also be possible to standardize the return types and enums to minimize duplication. I prefer the device-side return where possible, and we can add python-level logic to make sure we're not changing the API accidentally. You don't need to tackle that in this PR, I can work with it in a followup, but if not then please add a TODO for output format unification.

src/butina.cu
```diff
@@ -24,7 +24,7 @@
 
 /**
  * TODO: Future optimizations
```
scal444, line 28: Minimize the diff here please.

nvmolkit/clustering.py
```diff
@@ -157,76 +163,18 @@ def fused_butina(
...
-            end_idx = cluster_sizes[i + 1]
-            cluster_members = indices_cpu[start_idx:end_idx].tolist()
-
-            centroid = centroids[i]
-            members = [centroid] + [m for m in cluster_members if m != centroid]
-            clusters.append(tuple(members))
-        if return_centroids:
-            return clusters, cluster_sizes, centroids
-        return clusters, cluster_sizes
+        if not x.is_contiguous():
```
scal444: Small nitpick, you don't need the if statement, `x.contiguous()` is a zero-overhead op if already contiguous.

src/fused_butina.cu
```diff
@@ -0,0 +1,604 @@
...
+  candidate = cub::BlockReduce<std::uint64_t, kInitialArgMaxBlockSize>(storage).Reduce(candidate, cubMax());
+  if (threadIdx.x == 0) {
+    storeCandidate(candidate, maxValue, maxIndex);
+  }
+}
+
+// Compare the selected centroid against every active molecule and extract all of its neighbors into the current
+// cluster.
+template <FingerprintSimilarityMetric Metric>
+__global__ void extractClusterKernel(const cuda::std::span<const std::uint32_t> fingerprints,
```
scal444, line 259: Likely overkill, but we could reduce the global atomicAdd tension by precomputing warp and/or block level neighbor scans and adding multiple entries to global at a time. Worth adding a todo but not necessary here.

scal444 (commented):
Thanks for the contribution, I'll be sure to benchmark it!

My main concern is code duplication, but I know the fused/non-fused versions are subtly different enough so that consolidating all the kernels might not make sense. Could you generally take a look and see if anything is worth pulling out into a butina_common.cuh/.cu file?

src/fused_butina.h
Matthew-Neba: Done. I have also done a decent amount of simplification and deduping of the various butina implementations. There's still some stuff to do for the reordering=true butina path and I have left a TODO for that to be tackled in the future.

src/fused_butina.cu
Matthew-Neba: Thanks for the insight! Yeah, I was able to use warp level scans like you suggested. Should reduce the atomic adds by a factor of about 1/32

scal444:
Hi Matthew - I benchmarked the current PR on my local 5080. I'm seeing good speedups for the smaller cases but perf regressions as it grows.

```
   Molecules     Main avg.       PR avg.           PR result
  ━━━━━━━━━━━  ━━━━━━━━━━━━  ━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━
          1k      0.563 ms      0.158 ms        3.56× faster
  ───────────  ────────────  ────────────  ──────────────────
          5k      2.398 ms      1.596 ms        1.50× faster
  ───────────  ────────────  ────────────  ──────────────────
         10k      5.951 ms      5.815 ms    effectively tied
  ───────────  ────────────  ────────────  ──────────────────
         20k     17.824 ms     22.064 ms        1.24× slower
  ───────────  ────────────  ────────────  ──────────────────
         30k     31.562 ms     48.922 ms        1.55× slower
  ───────────  ────────────  ────────────  ──────────────────
         40k     48.762 ms     86.065 ms        1.77× slower
  ───────────  ────────────  ────────────  ──────────────────
        100k    226.471 ms    532.878 ms        2.35× slower
  ───────────  ────────────  ────────────  ──────────────────
        500k       4.788 s      27.581 s        5.76× slower

```

I traced it to the initial neighbors kernel (~95%+ of runtime). - it's scaling quadratically and is less efficient that the original triton. We may need to rework it to be more efficient. If you have time to take a look at it, that would be great, otherwise I can give it a stab later this week.

Matthew-Neba:
Ok, thanks for the benchmark results. I’ll take a crack at it today

docs/changelog.md
```diff
@@ -32,6 +32,9 @@ nvMolKit 0.5.0 adds three new GPU-accelerated APIs: Torsion Fingerprint Deviatio
 - Fix empty result handling in substructure search with `uniquify` when all matches were already unique ([#112](https://github.com/NVIDIA-Digital-Bio/nvMolKit/issues/112))
 
 ### Miscellaneous
+- `fused_butina()` now places its optional `stream` argument last; callers using positional arguments should pass
```
scal444: Again, should not mess with the changelog.

src/butina.h
```diff
@@ -16,66 +16,115 @@
 #ifndef NVMOLKIT_BUTINA_H
 #define NVMOLKIT_BUTINA_H
 
+#include <cuda_runtime.h>
+
+#include <cstdint>
+#include <cuda/std/span>
+
+#include "src/fingerprint_similarity_device.cuh"
```
scal444: I generally would like to avoid polluting `.h` with `.cuh` includes. Any `.h` file should be include-able by a `.cpp` file without worrying about if it's bringing in something the requires nvcc. Another thing we haven't been great about throughout the codebase, but I'm trying to enforce it going forward.

src/fused_butina.cu
```diff
@@ -0,0 +1,612 @@
...
+                                                 originalIndicesSpan.data(),
+                                                 sortedIndicesSpan.data(),
+                                                 n,
+                                                 0,
+                                                 sizeof(int) * 8,
+                                                 stream));
+  cudaCheckError(cudaMemsetAsync(neighborCountsSpan.data(), 0, neighborCountsSpan.size_bytes(), stream));
+  const int numInitialTiles = (n + kInitialTileSize - 1) / kInitialTileSize;
+  initialNeighborCountKernel<Metric>
+    <<<dim3(numInitialTiles + 1, (numInitialTiles + 1) / 2), kInitialBlockSize, 0, stream>>>(fingerprints,
```
scal444: I think we'll run into a grid limit crash here on large enough sizes. We had a similar issue with the fused version https://github.com/NVIDIA-BioNeMo/nvMolKit/pull/194

scal444 (changes requested):
Thanks for all the extra cleanup! Perf now looks great - 

```
   Size          Main    Previous PR    Updated PR         vs main
  ━━━━━━  ━━━━━━━━━━━━  ━━━━━━━━━━━━━  ━━━━━━━━━━━━  ━━━━━━━━━━━━━━
     1k      0.563 ms       0.158 ms      0.148 ms    3.81× faster
  ──────  ────────────  ─────────────  ────────────  ──────────────
     5k      2.398 ms       1.596 ms      0.453 ms    5.30× faster
  ──────  ────────────  ─────────────  ────────────  ──────────────
    10k      5.951 ms       5.815 ms      0.997 ms    5.97× faster
  ──────  ────────────  ─────────────  ────────────  ──────────────
    20k     17.824 ms      22.064 ms      2.905 ms    6.14× faster
  ──────  ────────────  ─────────────  ────────────  ──────────────
    30k     31.562 ms      48.922 ms      6.088 ms    5.18× faster
  ──────  ────────────  ─────────────  ────────────  ──────────────
    40k     48.762 ms      86.065 ms     10.312 ms    4.73× faster
  ──────  ────────────  ─────────────  ────────────  ──────────────
   100k    226.471 ms     532.878 ms     57.488 ms    3.94× faster
  ──────  ────────────  ─────────────  ────────────  ──────────────
   500k       4.788 s       27.581 s       1.531 s    3.13× faster

```

A few small comments then this can go in.

Matthew-Neba:
Apologies for messing around with the changelog, won't touch again. I was also under the impression that you can launch as many blocks as you want in CUDA and the CUDA scheduler will map those virtual blocks to real hardware, didn't know there was a limit. The other requests have also been fulfilled. Thanks!

src/fused_butina.cu
```diff
@@ -0,0 +1,601 @@
...
+                                                            int                  lhsBitCount,
+                                                            int                  rhsBitCount,
+                                                            int                  numWords,
+                                                            float                threshold) {
+  int intersection = 0;
+  for (int word = 0; word < numWords; ++word) {
+    intersection += __popc(lhs[word] & rhs[word]);
+  }
+
+  return intersectionMatches<Metric>(intersection, lhsBitCount, rhsBitCount, threshold);
```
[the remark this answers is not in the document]
scal444: @Matthew-Neba can you check out this comment and see if it's still applicable?

scal444:
Yeah the SM 90 thing is weird - greptile isn't the only model hallucinating that issue, I think it's probably due to some internet cache of one of our initial blog posts about it.

scal444 (approved):

scal444:
Thanks again! That triton dep was useful for our initial implementation but has caused a lot of trouble with environments, happy to see it go for now.

Matthew-Neba:
Glad to have been a help

PR #227, opened 2026-07-21
MCS PR 1 - Extract executor ring functionality from substructure search to be reused

scal444:


evasnow1992 (approved):
Changes look good to me!

PR #228, opened 2026-07-21
Autotuning and benchmarks for FIRE APIs

scal444:
PR 7 for FIRE


evasnow1992 (approved):
Changes look good to me!

PR #229, opened 2026-07-23
Primitives and device queues for MCS - MCS PR 2/N

scal444:
Added data structures that are MCS specific. We have a few differences from substruct search that required new graph representations.

Added a seed object, fMCS grows seeds from a pool. 

Added a queue to push/pop candidate seeds in-device.


evasnow1992 (approved):
Thank you! Changes look good to me.

PR #230, opened 2026-07-24
MCS graph construction utils

scal444:
We use a simpler graph representation than substruct, since we don't need repeated label/recurse cycles and the scope of matching is typically more limited.

PR 3 towards #221


evasnow1992 (approved):

scal444:
As discussed I'll wait until CI is green and rebase before merging.

PR #231, opened 2026-07-24
fMCS seed grow utilities

scal444:
Utils for getting new candidate bonds to add and pruning. fMCS has as funky extension mechanism, where it first tries to add all new bonds. Then tries each bond singly. Then, other combinations. 

PR for #221


src/mcs/fmcs_cuda/fmcs_grow.cuh
```diff
@@ -0,0 +1,208 @@
...
+    if (group.thread_rank() == 0) {
+      seedBeginGrowStepWithinThread(childWorkspace.seed);
+      seedAddNewBondWithinThread(childWorkspace.seed, bonds[i]);
+    }
+    group.sync();
+
+    // Try to extend the parent's recorded match by this single bond.
+    if (matchFn(childWorkspace)) {
+      childSink(childWorkspace);
+    }
```
evasnow1992, line 202: Should we set bonds[i].alive to false if it could not match? This may help if bonds.alive is used again in a later stage. Not necessary if we only reconstruct from childSink from now.

evasnow1992 (approved):
One minor comment. Changes look good to me. Thanks!

src/mcs/fmcs_cuda/fmcs_grow.cuh
scal444: Done

PR #232, opened 2026-07-24
Align nvMolKit skill with BioNeMo expected representation and add evals. 

scal444:
This will be consumed by https://github.com/NVIDIA-BioNeMo/bionemo-agent-toolkit


evasnow1992 (approved):

PR #235, opened 2026-07-27
Reject cross-device streams in acquireExternalStream

scal444:
acquireExternalStream validated a caller-supplied stream handle with cudaStreamQuery, which reports whether the stream is live, not which device owns it. CUDA permits querying a stream that belongs to another device in the same process, so a foreign stream returned cudaSuccess and was accepted. The cross-device rejection the function advertises was therefore never actually happening.

Resolve the stream's owning device and compare it against the current device. cudaStreamGetDevice does this directly but requires CUDA 12.8, above our supported floor, so below that fall back to probing with an event: an event and the stream it is recorded into must share a CUDA context, so recording an event created on the current device fails with cudaErrorInvalidResourceHandle when the stream belongs to another device.

The driver API (cuStreamGetCtx) was avoided deliberately: it changes signature in CUDA 13 and would add a libcuda link dependency.

DeviceTest.AcquireExternalStreamWrongDevice covers this but only runs with 2+ GPUs, which is why it went unnoticed until it failed in multi-GPU CI.

### DIFF
--- a/src/utils/device.cpp
+++ b/src/utils/device.cpp
@@ -37,18 +37,60 @@ WithDevice::~WithDevice() {
   cudaCheckErrorNoThrow(cudaSetDevice(original_device_id_));
 }
 
+namespace {
+
+//! Returns true if `stream` belongs to the current device.
+//! cudaStreamQuery accepts streams from any device in the process, so it cannot answer this on its
+//! own.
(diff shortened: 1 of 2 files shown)


evasnow1992 (approved):
Thank you for fixing the test failure.

PR #241, opened 2026-07-31
Final stretch of standardizing Butina

Matthew-Neba:
## Summary

  - Standardize Butina clustering behavior with RDKit.
  - Ensure deterministic higher-index tie-breaking and consistent cluster ordering (exactly matching RDkit), took a small hit in performance due to the extra work done

### DIFF
--- a/src/butina.cu
+++ b/src/butina.cu
@@ -29,12 +29,6 @@
  * - Keep a live list of active indices and only dispatch counts for those.
  */
 
-/**
- * TODO: Butina standardization (and make deterministic)
- * - Align the non-fused reordering=true path with the other butina implementations and RDKit:
- * 1) select the higher-index argmax (on ties) for each iteration
- * 2) and keep selection-order cluster IDs instead of renumbering clusters by final size.
- */
 namespace nvMolKit {
 
 namespace {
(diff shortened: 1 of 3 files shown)


src/butina.cu
```diff
@@ -638,107 +621,26 @@ __global__ void assignFixedOrderActiveRowKernel(const cuda::std::span<const uint
...
+    decodeAndStoreButinaCandidate(maxCandidate, outValue, outIndex);
   }
 }
-#endif  // CUB_VERSION < 200800
-
-//! Helper class to run ArgMax on device data.
-//! Uses CUB's DeviceReduce::ArgMax when available (CCCL >= 2.8.0), otherwise falls back to custom kernel.
-class ArgMaxRunner {
- public:
-  ArgMaxRunner([[maybe_unused]] size_t num_items, cudaStream_t stream)
```
scal444, line 680: This backup path was here because older versions of CCCL ship with olders versions of CUDA. The first CUDA that natively ships sufficient CCCL is 12.9 https://docs.nvidia.com/cuda/archive/12.9.1/cuda-toolkit-release-notes/index.html#cuda-toolkit-major-components

I'm hoping to support 12.5 for a few more release cycles due to some users being stuck on older AWS instances. Can you check on this? 

EDIT: I see your replacement is in-kernel cubmax, not devicemax - so it may be this concern is invalid. If you can quickly double-check the API is supported in older cccl, this is probably good to go. Thanks!

Matthew-Neba:
Yeah, I just double checked it on 12.5, it works. Since the O(n^2) distance thresholding step dominates the runtime anyways (>60%), and some added complexity would'hv been needed to make the new changes work with DeviceReduce (which would'hv improved runtime by 1.5% at 30k fingerprints), decided to go for the in kernel cubmax.

scal444 (approved):
Nice work again!

Matthew-Neba:
Thanks for taking the time to review these!

PR #242, opened 2026-07-31
Consolidate more benchmark code, make ff optimize more robust

scal444:
Consolidated CPU count with fallback for older python (< 3.13).

Added pre-run checks for MMFF/UFF forcefield compatibility so we don't choke on inputs with a few bad molecules.


evasnow1992 (approved):
Thank you for the consolidation. Changes are clean.

PR #244, opened 2026-07-31
DFS Substruct backend algorithm

scal444:
Up to 5x faster in kernel with up to 2x faster e2e substruct runtime. This will be the backend for MCS fallback match as well.


nvmolkit/substructure.cpp
```diff
@@ -92,6 +94,28 @@ void setGpuIdsPy(nvMolKit::SubstructSearchConfig& config, const object& iterable
...
 }
 
+std::string getAlgorithmPy(const nvMolKit::SubstructSearchConfig& config) {
+  switch (config.algorithm) {
+    case nvMolKit::SubstructAlgorithm::GSI:
+      return "gsi";
+    case nvMolKit::SubstructAlgorithm::DFS:
+      return "dfs";
+    case nvMolKit::SubstructAlgorithm::VF2:
+      return "vf2";
```
evasnow1992: Just to clarify, we don’t intend to expose VF2 through the Python frontend, right?

evasnow1992 (approved):
LGTM. Only one clarification question

nvmolkit/substructure.cpp
scal444: Good catch, changed that to throw an informative error.

PR #245, opened 2026-08-03
Extract the labelled-subgraph frontend from the substructure backend

scal444:
The per-pair table machinery -- back-edge extraction, edge-key classification into Uniform/Dual/General, the adjacency tables, and buildCandidates -- moves to src/subgraph/labeled_dfs.cuh, generic over an adapter that presents a query in search order.

Depth becomes a search position rather than a query atom index, so a frontend can search a subset of a query in a permuted order. The substructure backend supplies SubstructAdapter with the identity order and is otherwise unchanged.

This will allow us to drop in MCS-specific bits without major further changes to DFS

### DIFF
--- a/src/subgraph/candidate_tables.cuh
+++ b/src/subgraph/candidate_tables.cuh
@@ -1,0 +1,394 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


scal444:
Towards #221. @evasnow1992 this is mostly just code movement

src/subgraph/candidate_tables.cuh
```diff
@@ -0,0 +1,394 @@
...
+      continue;
+    }
+    seen |= 1ULL << neighborDepth;
+    const int      mapped = mapping[neighborDepth];
+    const uint32_t key    = adapter.edgeKeyAt(depth, slot);
+    if constexpr (Adapter::kCachesTargetRows) {
+      candidates.andEq(adapter.neighborsMatchingPacked(shared.targetAdjacency[warp][mapped].packedRow,
+                                                       shared.targetAdjacencyAlt[warp][mapped].packedRow,
+                                                       key));
+    } else {
```
evasnow1992, line 384: Just to double confirm, this false branch is currently unreachable as SubstructAdapter set it true, but it will be used in a follo-up PR by MCS right?

evasnow1992 (approved):
One clarification comment. Changes look good to me. Thank you!

src/subgraph/candidate_tables.cuh
scal444: Correct!

PR #249, opened 2026-08-04
Add fMCS search support primitives

scal444:
Per-pair kernel inputs and results, cooperative incumbent tracking, sorted seed-queue insertion, the RDKit-parity remaining-size bound, and checkSeedMatchAndAppendCooperative, which runs the greedy extender and falls back to the exact substructure search.


evasnow1992 (approved):

PR #250, opened 2026-08-05
Add CUDA fMCS engine

scal444:
Ports the tier-dispatched fMCS driver and block kernel onto the shared subgraph candidate tables. The seed/substructure match caches are gone: the kernel calls the fallback matcher directly, so the cache storage parameter, its device allocation, and the global partial-storage slab all drop out.

Tier-128 now needs global substructure scratch at block size 256 as well as 512, since the shared per-group scratch no longer fits.

Covered by two suites: test_fmcs_core for result correctness (paths, trees, cycles, molecules, degenerate and disconnected inputs) and test_fmcs_engine_validation for the batch machinery (tier and block-size dispatch, batch chunking, multi-executor runs, timeout and overflow reporting, and atom/bond mapping consistency).

### DIFF
--- a/src/mcs/fmcs_cuda/fmcs.cu
+++ b/src/mcs/fmcs_cuda/fmcs.cu
@@ -1,0 +1,629 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <cooperative_groups.h>
+#include <cuda_runtime.h>
+
+#include <algorithm>
+#include <array>
+#include <cstdint>
+#include <memory>
+#include <stdexcept>
+#include <string>
+#include <type_traits>
+#include <utility>
+#include <variant>
+#include <vector>
+
+#include "src/mcs/fmcs_cuda/fmcs.cuh"
+#include "src/mcs/fmcs_cuda/fmcs_kernel.cuh"
+#include "src/mcs/fmcs_cuda/fmcs_match_tables.cuh"
+#include "src/mcs/fmcs_cuda/fmcs_policy.cuh"
+#include "src/utils/device.h"
+#include "src/utils/gpu_executor_ring.h"
+
+namespace mcs {
+namespace fmcs {
+
+namespace {
+
+void checkCuda(cudaError_t err, const char* context) {
+  if (err != cudaSuccess) {
+    throw std::runtime_error(std::string("fMCS dispatch CUDA error at ") + context + ": " + cudaGetErrorString(err));
+  }
+}
+
+/// uint32 host-side mirror of a CSR @ref Graph plus packed bond metadata
+/// matching @ref enumerateBonds ordering.
+struct PackedGraphHost {
+  std::vector<uint32_t> rowOffsets;
+  std::vector<uint32_t> colIndices;
+  /// Per-adjacency entry bond id, parallel to @c colIndices.
+  std::vector<uint32_t> bondIndices;
+  std::vector<uint32_t> bondEndpoints;  // (u << 16) | v, u < v
+};
+
+PackedGraphHost packGraph(const Graph& g) {
+  constexpr uint32_t kUnsetBondIndex = 0xFFFFFFFFu;
+  PackedGraphHost    out;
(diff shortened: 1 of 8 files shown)


src/mcs/fmcs_cuda/fmcs_kernel.cuh
```diff
@@ -0,0 +1,548 @@
...
+      if (aliveOverflow)
+        break;
+      if (aliveCount > 1) {
+        unsigned int erasedCount = 0;
+        if (groupRank == 0) {
+          erasedCount = static_cast<unsigned int>(newBondCount[groupId] - aliveCount);
+        }
+        erasedCount                             = group.shfl(erasedCount, 0);
+        const unsigned long long maxComposition = (1ULL << aliveCount) - 1ULL;
+        for (unsigned long long composition = maxComposition; composition != 0ULL; --composition) {
```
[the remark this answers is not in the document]
scal444: For a molecule is this actually a problem? While some aspects of fmcs are generic I don't really care about edge cases in generic graphs

src/mcs/fmcs_cuda/fmcs.cuh
```diff
@@ -0,0 +1,87 @@
...
+  int    blockSize          = 128;
+  /// Per-pair wall timeout in milliseconds.  0 = no timeout.
+  float  timeoutMs          = 0;
+  /// Max pairs per kernel launch in the batch API.  0 = auto from GPU memory.
+  int    batchSize          = 0;
+  /// Number of asynchronous executor streams for tier sub-batches. 1 = serial.
+  int    executorsPerRunner = 1;
+  /// Optional absolute wall deadline in seconds since epoch; pairs not
+  /// started by this time are reported as timed out.  0 = disabled.
+  double wallDeadlineSec    = 0;
```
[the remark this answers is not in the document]
scal444: Removed thanks

src/mcs/fmcs_cuda/fmcs.cu
```diff
@@ -0,0 +1,629 @@
...
+  out.colIndices.reserve(g.colIndices.size());
+  for (size_t v : g.colIndices)
+    out.colIndices.push_back(static_cast<uint32_t>(v));
+  out.bondIndices.assign(g.colIndices.size(), kUnsetBondIndex);
+  out.bondEndpoints.reserve(static_cast<size_t>(g.numEdges));
+  for (int u = 0; u < g.numVertices; ++u) {
+    const size_t begin  = g.rowOffsets[u];
+    const size_t end    = g.rowOffsets[u + 1];
+    const size_t degree = end - begin;
+    if (degree > static_cast<size_t>(kMaxNeighborsPerAtom)) {
```
scal444, line 75: @evasnow1992 limit added here for 8 neighbors

evasnow1992: Thank you!

src/mcs/fmcs_cuda/fmcs.cuh
```diff
@@ -0,0 +1,88 @@
...
+/// The objective (connected MCES, MaximizeBonds, Threshold=1) is baked
+/// into the algorithm itself and is not exposed as a knob.
+struct Parameters {
+  /// CUDA block size for the per-pair kernel. Supported: 64, 128, 256, 512.
+  /// Tier-128 searches at 512 threads place substructure scratch in global
+  /// memory; smaller tiers retain shared-memory scratch.
+  int   blockSize          = 128;
+  /// Per-pair wall timeout in milliseconds.  0 = no timeout.
+  float timeoutMs          = 0;
+  /// Max pairs per kernel launch in the batch API.  0 = auto from GPU memory.
```
evasnow1992: This comment appears misleading. When `batchSize == 0`, the chunk size is set on line 582 of `src/mcs/fmcs_cuda/fmcs.cu` as:

`
const size_t chunkSize =
    params.batchSize > 0 ? static_cast<size_t>(params.batchSize) : N;
`

Here, `N = a.size()` is the total number of input pairs, so the value is not automatically derived from available GPU memory.

src/mcs/fmcs_cuda/fmcs.cu
```diff
@@ -0,0 +1,629 @@
...
+  size_t              csrBufferBytes  = 0;
+  DevicePerPairInput* dPairInputs     = nullptr;
+  void*               dResultsBuffer  = nullptr;
+  void*               dQueueStorage   = nullptr;
+  void*               dScratchStorage = nullptr;
+};
+
+constexpr int kMaxFmcsExecutorsPerRunner = 8;
+
+struct FmcsExecutor {
```
evasnow1992, line 179: Should we add a destructor that calls freeBatchBuffers() to ensure the allocated memory is released if checkCuda throws at the queue allocation (line 326)? Or add a call to freeBatchBuffers() after line 326?

evasnow1992 (approved):
Two minor comments. Changes look good to me.

src/mcs/fmcs_cuda/fmcs.cu
scal444: We shouldn't have had this in the first place. I could have sworn I removed all of the manual device memory management but it must have snuck back in during a rebase. Changed everything to AsyncDeviceVector so it's natively handled in the destructor. Thanks for pointing it out

src/mcs/fmcs_cuda/fmcs.cuh
scal444: Fixed the comment thanks

PR #252, opened 2026-08-06
Standardize Python benchmark infrastructure

scal444:
* Removed last bespoke timing code
* Standardized --no-rdkit, --no-nvmolkit flags
* Standardized multiprocess rdkit with timeout + partial results collection
* Standardize output writing to csvs

Note that none of this is being applied to the similarity bench - it's a weird edge case that uses pyperf, we'll leave it alone.


evasnow1992 (approved):
Thank you for the consolidation. Changes look good to me.

PR #253, opened 2026-08-06
Remove RDKit SubstructLibrary benchmark mode

scal444:
This was always slower than our hand-rolled version. Might be something we could help out rdkit with in the future.

### DIFF
--- a/benchmarks/substruct_bench.py
+++ b/benchmarks/substruct_bench.py
@@ -17,16 +17,12 @@
 
 Compares two approaches:
   1. nvmolkit GPU-accelerated substructure search
-  2. RDKit SubstructMatch API (raw or SubstructLibrary mode)
+  2. RDKit direct SubstructMatch API with optional multiprocessing
 
 Supports three search modes:
   - hasSubstructMatch: Boolean match detection (faster)
   - countSubstructMatches: Count of matches per target/query pair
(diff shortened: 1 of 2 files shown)


benchmarks/substruct_bench.py
```diff
@@ -883,7 +764,8 @@ def main():
             num_gpus = applied_num_gpus if name == "nvmolkit" else config_row["num_gpus"]
             nvmolkit_config_source = config_source if name == "nvmolkit" else "N/A"
             if is_rdkit:
-                rdkit_match_mode, rdkit_threads = rdkit_variant_meta[name]
+                rdkit_match_mode = "raw"
+                rdkit_threads = rdkit_variant_meta[name]
                 rdkit_max_seconds = args.rdkit_max_seconds
             else:
                 rdkit_match_mode = "N/A"
```
evasnow1992: I assume we're still keeping the hardcoded rdkit_match_mode column for schema compatibility with previous results. Otherwise, functionally it looks vestigial.

evasnow1992 (approved):
One minor comment, non-blocker. Changes look good to me.

benchmarks/substruct_bench.py
scal444: Done

scal444: (as in, removed the field and relevant variables)

PR #254, opened 2026-08-06
Make TFD benchmark failures explicit

scal444:
Failure cases weren't actually propagating correctly and csv rows could be incorrectly written.

### DIFF
--- a/benchmarks/tfd_bench.py
+++ b/benchmarks/tfd_bench.py
@@ -54,6 +54,7 @@ from rdkit.Chem import TorsionFingerprints
 
 import nvmolkit.tfd as nvmol_tfd
 
+
 def generate_conformers_batch(
     mols: List[Chem.Mol],
     num_confs: int,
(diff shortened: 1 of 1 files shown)


benchmarks/tfd_bench.py
```diff
@@ -82,50 +81,25 @@ def generate_conformers_batch(
...
     max_mols: int = 100,
-    smiles_file: str | None = None,
     num_workers: int = 0,
     seed: int = 42,
 ) -> List[Chem.Mol]:
-    """Prepare molecules with conformers, using precomputed pickle if available.
+    """Prepare molecules with conformers from the explicitly supplied inputs.
 
     Args:
         input_mols: Parsed RDKit molecules (used when no precomputed pickle is found).
```
evasnow1992: This comment is no longer accurate after this PR.

evasnow1992 (approved):
One stale argument docstring. Changes look good to me.

benchmarks/tfd_bench.py
scal444: Fixed

PR #255, opened 2026-08-06
Remove synthetic Butina benchmark padding

scal444:
This was a bad idea from me in the first place - synthetic randomized data doesn't cluster like normal molecules and could give misleading results. We've got big enough datasets, we will just use them when we want to push the envelope


evasnow1992 (approved):
Thank you for the clean-up.

PR #257, opened 2026-08-07
MCS RDKit support

scal444:
Orchestration and labeling support for various rdkit options.

fMCS has a lot of toggles we don't yet support - for clarity, these are included in our configuration structures but all throw with informative errors.


evasnow1992 (approved):
Thank you for adding the wrappers. Changes look good to me.

PR #259, opened 2026-08-10
Fix MMFF parametrization of cyclophosphazine

### LINKED ISSUE #258: MMFF fails to parametrize cyclophosphazine
We're mirroring an RDKit bug here https://github.com/rdkit/rdkit/issues/9473

scal444:
Fixes #258


evasnow1992 (approved):
Thank you for the fix. LGTM.

PR #261, opened 2026-08-11
Host and device memory improvements for MCS

scal444:
* Packs multiple data into a single buffer to reduce the number of memcpys
* Uses pinned host memory
* Persistent device buffers between batches

Together, this mostly removes any memory-based overhead seen on Nsight Systems profiles


evasnow1992 (approved):
Thank you for this performance improvement change.

PR #263, opened 2026-08-12
Reduce max executors and improve batch size coverage in mcs

scal444:


evasnow1992 (approved):
This current PR may have conflicts with #262 (blockSize is removed in #262). The changes are clean assuming that 262 will be merged first, and the current one will be rebased on it.

PR #270, opened 2026-08-19
Add missing docs for MCS

scal444:


docs/api/nvmolkit.rst
```diff
@@ -80,6 +80,24 @@ Butina Clustering
    :template: function_template.rst
 
    clustering.butina
+   clustering.fused_butina
```
evasnow1992, line 83: Since we're exposing fused_butina here, the landing-page Butina bullet (index.rst:184) is now incomplete and may need an update.

evasnow1992 (approved):
One minor comment. LGTM.

docs/api/nvmolkit.rst
scal444: Done

PR #271, opened 2026-08-19
Update nvmolkit skill for v0.6 features

scal444:


skills/nvmolkit-usage/evals/evals.json
```diff
@@ -168,6 +168,21 @@
...
       "expected_script": null
+    },
+    {
+      "id": "nvmolkit-usage-012",
+      "prompt": "Write nvMolKit code that finds maximum common substructures for the explicit molecule pairs [(0, 1), (2, 3), (0, 2)]. Preserve that pair order, use element and bond-order matching, set a 5-second timeout for each pair, and report atom/bond counts plus whether each search timed out.",
+      "expected_output": "The agent constructs RDKit molecules, calls nvmolkit.mcs.findMCS once in explicit-pairs mode, and reads the ordered MCSBatchResult correctly, including each MCSResult's counts and canceled flag.",
+      "assertions": [
+        "The code imports findMCS from nvmolkit.mcs and calls it once for the batch",
+        "The code passes mode='pairs' and the explicit pairs [(0, 1), (2, 3), (0, 2)] without reordering or deduplicating them",
+        "The code uses atom_compare='elements', bond_compare='order', and timeout_seconds=5",
```
evasnow1992: The prompt asks for "use element and bond-order matching" which are actually the findMCS defaults. An agent that somehow identifies the defaults and passes only timeout_seconds=5 produces identical behavoir but then fails this assertion. May consider switching to non-default values in the prompt then assert on them. Not a blocker though.

evasnow1992 (approved):
LGTM.

skills/nvmolkit-usage/evals/evals.json
scal444: Fixed

PR #278, opened 2026-08-21
Fix benchmark loader progress

scal444:
These were showing < 1% completeness even though they were progressing, replaced the implementation.


evasnow1992 (approved):
Thank you for fixing this progress bar issue. Changes look good to me.

PR #280, opened 2026-08-21
Fix SMARTS hydrogen counts for explicit neighbors

### LINKED ISSUE #276: SMARTS hydrogen counts exclude explicit hydrogen neighbors
AtomHCount implementation does not count explicit hydrogen neighbors, including deuterium, while RDKit does. Queries such as [CH3], [CH2], and [cH] can disagree.

```python
  from rdkit import Chem
  from nvmolkit.substructure import getSubstructMatches

  target = Chem.MolFromSmiles("[2H]C([2H])([2H])N")
  query = Chem.MolFromSmarts("[CH3]")

  print(target.HasSubstructMatch(query))                    # True
  print(bool(getSubstructMatches([target], [query])[0][0])) # False
```

scal444:
Fixes #276

### DIFF
--- a/nvmolkit/tests/test_substructure.py
+++ b/nvmolkit/tests/test_substructure.py
@@ -15,7 +15,6 @@
 
 """Tests for GPU-accelerated substructure search."""
 
-import os
 from dataclasses import dataclass
 from pathlib import Path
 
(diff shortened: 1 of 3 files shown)


evasnow1992 (approved):
Thank you for fixing this issue.

PR #281, opened 2026-08-21
Add conformer RMSD alignment modes

scal444:
Distinguishes between two different common RDKit RMSD paths - the conformer RMSD matrix we were allegedly replacing in the first place aligns on a single conformer, whereas a true pairwise RMSD aligns individually.

Since the previous RMSD implementation worked by computing RMSD without **actually** aligning mols, we needed significant code to match. We could have done this via the original kabsch mechanism, but would give up an O(n) for O(n^2) path for the align on 0 mechanism.


evasnow1992 (approved):
Changes look good for me. Thank you for clearly distinguish the two modes.

PR #282, opened 2026-08-21
Avoid flaky authenticated Git fetch for nanobench

scal444:
- fetch nanobench v4.1.0 from a checksummed source archive instead of a nested Git clone
- disable persisted checkout credentials in the C++ build workflow, which does not perform later authenticated Git operations

The C++ configure job intermittently receives an authentication challenge while cloning the public nanobench repository. The checkout action persists a GitHub-wide authorization header in the checkout Git config, and CMake runs the dependency clone beneath that checkout. Using an archive removes the Git smart-HTTP and credential-inheritance path entirely; disabling credential persistence prevents similar unintended authentication in later commands.

### DIFF
--- a/.github/workflows/cpp-build-test.yml
+++ b/.github/workflows/cpp-build-test.yml
@@ -47,6 +47,8 @@ jobs:
 
       - name: Check out source tree
         uses: actions/checkout@v4
+        with:
+          persist-credentials: false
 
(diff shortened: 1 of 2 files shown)


evasnow1992 (approved):
Thank you for fixing this random credential issue.

PR #284, opened 2026-09-01
Optimize Tanimoto similarity for skinny queries

scal444:
Optimizes small dimension queries by transposing the small dim and using optimal matrix dimensions.  N=1 is still inefficient, but much less so.

### DIFF
--- a/src/similarity_kernels.cu
+++ b/src/similarity_kernels.cu
@@ -120,7 +120,8 @@ template <FingerprintSimilarityMetric similarityType,
           size_t BLOCK_TILE_SIZE_Y,
           size_t BLOCK_TILE_SIZE_K,
           size_t NUM_WARP_X,
-          size_t NUM_WARP_Y>
+          size_t NUM_WARP_Y,
+          bool   TRANSPOSE_OUTPUT = false>
 __device__ void crossSimilarityKernelTensorOp(const cuda::std::span<const kThreadReductionType> bitsOne,
                                               const cuda::std::span<const kThreadReductionType> bitsTwo,
                                               const size_t                                      elementsPerMolecule,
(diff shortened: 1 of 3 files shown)


src/similarity_kernels.cu
moradza: `TRANSPOSE_OUTPUT` makes the global writes strided, especially for the 16×64 and 32×64 tiles. Could we traverse C_smem column-wise in this branch so each warp writes 32 consecutive output values? The existing +1 padding should keep the shared-memory reads conflict-free.
```
const int smem_row = TRANSPOSE_OUTPUT
                           ? linear_idx % BLOCK_TILE_SIZE_Y
                           : linear_idx / BLOCK_TILE_SIZE_X;
const int smem_col = TRANSPOSE_OUTPUT
                           ? linear_idx / BLOCK_TILE_SIZE_Y
                           : linear_idx % BLOCK_TILE_SIZE_X;
```

src/similarity_kernels.cu
```diff
@@ -514,33 +550,17 @@ void launchCrossTanimotoSimilarity(const cuda::std::span<const std::uint32_t> bi
   size_t n = bitsTwo.size() / numBitsPerMolecule;
 
   if (isTensorOpsSupportedCached()) {
-    constexpr unsigned int BLOCK_TILE_SIZE_X{64U};
-    constexpr unsigned int BLOCK_TILE_SIZE_Y{64U};
-
-    constexpr unsigned int BLOCK_TILE_SIZE_K{32U};
-
-    constexpr int SMEM_IS_ENOUGH = (2 * (BLOCK_TILE_SIZE_X + BLOCK_TILE_SIZE_Y) * (BLOCK_TILE_SIZE_K + 1)) /
```
evasnow1992, line 522: Should we add a similar guard in launchTanimotoTensorKernel above?

evasnow1992 (approved):
Changes look good to me. One comment on whether it would be worth adding back a tile-size check in the new kernel launcher.

src/similarity_kernels.cu
scal444: Done

Issue #269, opened 2026-08-19
Undocumented v0.6.0 butina API break


scal444:
We removed the cluster_sizes return value from fused butina, breaking existing users migrating from v0.5 to v0.6.

```
 clusters, _, centroids = fused_butina(
             fingerprint_matrix, cutoff=distance_cutoff, return_centroids=True
   )
```
would pass on v0.5 but needs to be
```
clusters, centroids = fused_butina(
             fingerprint_matrix, cutoff=distance_cutoff, return_centroids=True
   )
```

It also changed to match the butina non-fused GPU output, so a user needs to re-form their downstream usage.

I think we should backfix this in v0.6.1 to provide user options for all the output modes

scal444:
After looking at this for a while, I think we need to bite the bullet and standardize the output formats rather than try to maintain backwards compatibility, because of the history of how the two API's return types have changed. Since we're doing it anyway, we should be GPU-native by default.

So, we'll have 2 output modes, GPU and RDKit-matching. RDKit mode will have centroids first, and counts are easily reconstructed, so will remove those custom output toggles - that was part of the v0.6 API break in the first place.

PR #285, opened 2026-09-01
Unify Butina outputs for device and RDKit compatibility mode. 

scal444:
Addresses #269 , though with another API modification. There's no good way to unify fused/nonfused Butina + our previous history of return types without a clean approach.

### DIFF
--- a/nvmolkit/clustering.py
+++ b/nvmolkit/clustering.py
@@ -45,12 +45,10 @@ _RDKitClusters = tuple[tuple[int, ...], ...]
 
 
 class ButinaOutput(Enum):
-    """Selects how Butina clustering results are returned.
+    """Output format for :func:`butina` and :func:`fused_butina`.
 
-    - ``RDKIT`` returns the same tuple-of-tuples representation as RDKit's
-      ``Butina.ClusterData``. The centroid is the first item in each cluster.
-    - ``DEVICE`` keeps the result on the GPU and returns a
-      :class:`ButinaDeviceResult`.
+    ``RDKIT`` returns the same tuple of clusters as RDKit's
+    ``Butina.ClusterData``. ``DEVICE`` returns a :class:`ButinaDeviceResult`.
     """
 
     RDKIT = "rdkit"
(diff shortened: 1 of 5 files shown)


skills/nvmolkit-usage/SKILL.md
evasnow1992: Also need import ButinaOutputMode here

nvmolkit/clustering.py
```diff
@@ -36,12 +41,69 @@
...
+
 
-def _wrap_result(result, return_centroids: bool):
-    if return_centroids:
-        cluster_ids, centroids = result
-        return AsyncGpuResult(cluster_ids), AsyncGpuResult(centroids)
-    return AsyncGpuResult(result)
+def _wrap_device_result(result) -> ButinaDeviceResult:
+    cluster_ids, centroids = _wrap_cluster_arrays(result)
+    cluster_sizes = torch.bincount(
```
evasnow1992: Would `torch.bincount` force a synchronization here?

If I’m not mistaken, the number of clusters is already known from `centroids.numel()`. In that case, we may be able to avoid the synchronization with something like:

```python
n = centroids.torch().numel()

ids = cluster_ids.torch().to(torch.int64)

cluster_sizes = torch.zeros(n, dtype=torch.int64, device=ids.device)
cluster_sizes.index_add_(0, ids, torch.ones_like(ids))
```

nvmolkit/clustering.py
```diff
@@ -36,12 +41,69 @@
...
+    return ButinaDeviceResult(cluster_ids, centroids, AsyncGpuResult(cluster_sizes))
+
+
+def _to_rdkit_clusters(cluster_ids: AsyncGpuResult, centroids: AsyncGpuResult) -> _RDKitClusters:
+    cluster_ids_array = cluster_ids.numpy()
+    centroids_array = centroids.numpy()
+    clusters = []
+    for cluster_id, centroid_value in enumerate(centroids_array):
+        centroid = int(centroid_value)
+        members = np.flatnonzero(cluster_ids_array == cluster_id)
```
evasnow1992: This appears to require one full pass over `cluster_ids` for each cluster, so it could degrade toward O(N²) when there are many singleton clusters. Would it be more efficient to sort once with `argsort` and then group/split the sorted indices (e.g., with `np.searchsorted`)?

skills/nvmolkit-usage/SKILL.md
scal444: Done

nvmolkit/clustering.py
scal444: Done

nvmolkit/clustering.py
scal444: Done - performance is not so important for the wrapper, but might as well.

evasnow1992 (approved):
Thank you for consolidating butina outputs and addressing the comments. Changes look good to me.

Issue #151, opened 2026-04-30
Add shellcheck to CI


scal444:
We have a lot of bash scripts now, some QA would be useful

PR #289, opened 2026-09-09
Add ShellCheck to CI

scal444:
Implements #151

### DIFF
--- a/admin/run_clang_tidy.sh
+++ b/admin/run_clang_tidy.sh
@@ -41,8 +41,9 @@ cd clang_tidy_build
 
 #clang-tidy-15  --config-file="$ROOT"/.clang-tidy  -p clang_tidy_build
 
-LLVM_SYMBOLIZER_PATH=$(command -v llvm-symbolizer-17)
-export LLVM_SYMBOLIZER_PATH
+if LLVM_SYMBOLIZER_PATH=$(command -v llvm-symbolizer-17); then
+  export LLVM_SYMBOLIZER_PATH
+fi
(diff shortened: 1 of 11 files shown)


admin/run_clang_tidy.sh
```diff
@@ -33,15 +33,16 @@ while getopts ":i" opt; do
...
 
 mkdir -p clang_tidy_build
 cd clang_tidy_build
 #CC=clang-15 CXX=clang++-15 cmake "$ROOT" -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DCMAKE_PREFIX_PATH="$RDKIT_PATH" -DNVMOLKIT_BUILD_TESTS=OFF -DNVMOLKIT_BUILD_BENCHMARKS=OFF
 
 #clang-tidy-15  --config-file="$ROOT"/.clang-tidy  -p clang_tidy_build
 
-export LLVM_SYMBOLIZER_PATH=$(which llvm-symbolizer-17)
+LLVM_SYMBOLIZER_PATH=$(command -v llvm-symbolizer-17)
+export LLVM_SYMBOLIZER_PATH
```
evasnow1992: With set -e on line 20, this now aborts the whole script if llvm-symbolizer-17 isn't installed. Looks like LLVM_SYMBOLIZER_PATH isn't read anywhere in this script. Do we really require it to be installed?

scal444: Fixed. It's optional

evasnow1992 (approved):
Changes look good to me.

Issue #288, opened 2026-09-03
Eigenvalue path for initial ETKDG coordinate gen


scal444:
We had prototypes for this years ago but have only supported useRandomCoords=True. I got a request for this at ACS, should not be difficult to support, but we need to be careful about our RNG and flaky tests.

PR #292, opened 2026-09-10
Enable eigenvalue ETKDG coordinate generation

scal444:
Match RDKit eigen initialization and retry behavior, make the batched eigensolver padding- and stream-aware, and enable the default EmbedMolecules path with updated docs and tests.\n\nCloses #288

### DIFF
--- a/src/forcefields/coord_gen.cu
+++ b/src/forcefields/coord_gen.cu
@@ -1,4 +1,4 @@
-// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
--- a/src/forcefields/coord_gen.cu
+++ b/src/forcefields/coord_gen.cu
@@ -32,6 +32,16 @@ namespace nvMolKit {
 
 namespace {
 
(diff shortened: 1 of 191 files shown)


src/forcefields/coord_gen.cu
```diff
@@ -80,49 +88,71 @@ __global__ void projectDistancesToPositions(const int      matrixDim,
...
-        const int batchIdx = idx / numEigs;
-        passFail[batchIdx] = 0;
+  if (systemIdx < numSystems) {
+    if (active != nullptr && active[systemIdx] == 0) {
+      passFail[systemIdx] = 0;
+      return;
+    }
+    const int    matrixDim     = matrixDimensions[systemIdx];
+    const int    systemNumEigs = min(numEigs, min(matrixDim, coordinateDimensions[systemIdx]));
+    bool         passed        = validDistanceMatrices[systemIdx] != 0 && eigensolverConverged[systemIdx] != 0;
```
evasnow1992: Adding `eigensolverConverged[systemIdx] != 0` here will cause single-atom molecules to fail while rdkit would succeed and place the atom at the origin.
Is this divergent behavior intended? If not, we may consider updating it to:
```cpp
bool passed = validDistanceMatrices[systemIdx] != 0 &&
              (eigensolverConverged[systemIdx] != 0 || matrixDim <= 3);
```
This would be more consistent with RDKit's behavior for `N <= 3`, where the zero-eigenvalue failure check is exempted. In that case, we may also need to update `EigensolverFailureFailsSmallSystems`, which currently asserts that `matrixDim` 1, 2, and 3 all fail.

scal444: Fixed

evasnow1992 (approved):
Thank you for addressing the comment. Changes look good to me.

Issue #267, opened 2026-08-19
Further CPU optimizations for fingerprint preprocessing


scal444:
We're very CPU bottlenecked on powerful GPUs, I've done some work optimizing the CPU stages, but we should take one more look at it. It may be that the RDKit APIs are a bottleneck we can't crack.

PR #297, opened 2026-09-10
Reduce python transfer overhead in morgan fingerprints

scal444:
The boost::python::extract is very slow. Added some more nvtx.

PR 5/5 for #267


evasnow1992 (approved):
Changes look good to me.

PR #298, opened 2026-09-10
Use mamba for Python CI dependencies

scal444:
This should help the 10 minute python test setup time


evasnow1992 (approved):
Changes look good to me.

PR #299, opened 2026-09-10
Fix concurrent ETKDG retry dispatch

### LINKED ISSUE #300: ETKDG can exit early on failing molecules without full maxAttempts exercised
A minor logical race leads to a slightly higher rate of failures in our version compared to rdkit for problematic molecules

scal444:
An ETKDG worker can temporarily run out of dispatchable work while another
  worker still has attempts in flight. The first worker currently treats that
  empty dispatch as global completion. If any in-flight attempts subsequently
  fail, their retries are never scheduled.

  Track in-flight attempts in the scheduler and make production dispatch wait
  until those attempts are recorded. Retry rounds now advance only after the
  current round has completed. Cancellation wakes waiting workers, and exceptions
  from dispatch threads are rethrown by `embedMolecules()` instead of being
  silently discarded.

  The scheduler also returns stable, per-molecule attempt IDs. This lets callers
  vary retry initialization independently of thread scheduling and batch order.
  
  Fixes #300

### DIFF
--- a/src/etkdg_impl.cpp
+++ b/src/etkdg_impl.cpp
@@ -281,33 +281,65 @@ Scheduler::Scheduler(const int numUniqueMols, const int numConfsPerMol, const in
   maxTriesPerMolecule_ = maxIterations_ * numConfsPerMol_;
   completedConformers_.resize(numUniqueMols, 0);
   totalAttempts_.resize(numUniqueMols, 0);
+  attemptsInFlightByMolecule_.resize(numUniqueMols, 0);
 }
 
-std::vector<int> Scheduler::dispatch(const int batchSize) {
-  std::vector<int> molIds;
-  molIds.reserve(batchSize);
-  const std::lock_guard lock(mutex_);
-  size_t                prevSize = 1;  // Just don't set to 0 to avoid loop iter 0 exit.
-  while (molIds.size() < batchSize && prevSize != molIds.size()) {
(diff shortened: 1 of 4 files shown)


src/etkdg_impl.cpp
```diff
@@ -281,48 +281,87 @@ Scheduler::Scheduler(const int numUniqueMols, const int numConfsPerMol, const in
...
+    attemptIds->clear();
+    attemptIds->reserve(batchSize);
+  }
+  for (size_t i = 0; i < numUniqueMolecules_; i++) {
+    // Keep one attempt in flight for each conformer still needed. Recording a
+    // failure immediately reopens its slot without waiting on other molecules.
+    while (completedConformers_[i] + attemptsInFlightByMolecule_[i] < numConfsPerMol_ &&
+           totalAttempts_[i] < maxTriesPerMolecule_) {
+      if (static_cast<int>(molIds.size()) >= batchSize) {
+        break;
```
evasnow1992: This break only stops the inner while loop. The outer for loop continues for all the remaining molecules. Minor one. Pattern existing before but impact may be larger now as we allow more dispatches.

evasnow1992 (approved):
Changes look good to me. One minor comment but not blocking.

src/etkdg_impl.cpp
scal444: Fixed thanks

PR #308, opened 2026-09-14
Add clang-tidy correctness CI

scal444:
Updates to use clang-tidy 22 which has limited CUDA support. 

We're starting with correctness-only items, then will expand with coverage and fixes .

Only true bug fixed in this diff is an unlikely overrun adding 1 to maxBatchSize BEFORE converting to size_t instead of after. The rest are defensive, or needed to get Clang CUDA compiling.


evasnow1992 (approved):
Thank you adding this additional gate. Changes look good to me.

PR #309, opened 2026-09-18
Enable developer warnings in C++ CI

scal444:


evasnow1992 (approved):
Thank you for switching on the warnings in C++ CI. Changes look good to me.

PR #311, opened 2026-09-18
Expand ruff checks, and apply non-formatting checks to CI

scal444:


nvmolkit/types.py
```diff
@@ -21,9 +21,13 @@
 import numpy as np
 import torch
 
-from nvmolkit import _arrayHelpers  # noqa: F401
-from nvmolkit import _embedMolecules  # type: ignore
-from nvmolkit import _types
+# Import order is significant: these native modules register converters used by
+# the modules imported after them.
+from nvmolkit import (
```
evasnow1992: Should we wrap this block with  `# isort: off` and  `# isort: on`?
The ruff's I rules this PR enforces will sort the names inside the parenthesised import alphabetically.

evasnow1992 (approved):
Changes look good to me. One non-blocker comment.

nvmolkit/types.py
scal444: Unfortunately yes I think so. It's a consequence of boost.python. We'll take another look at it when the nanobind transition is done. Thanks for pointing it out.

PR #313, opened 2026-09-18
Deduplicate Recursive SMARTS Solves

scal444:
When the same recursive SMARTS occurs multiple times at the same level, we should solve it once and broadcast rather than solving multiple times. We do that simply by mapping the pattern to multiple output bits. If that ever becomes a problem we could alias bits.


src/substruct/recursive_preprocessor.cu
```diff
@@ -315,9 +333,9 @@ void RecursivePatternPreprocessor::preprocessMiniBatch(
   cudaCheckError(cudaGetLastError());
 }
 
-void preprocessRecursiveSmarts(SubstructTemplateConfig           templateConfig,
-                               const MoleculesDevice&            targetsDevice,
-                               const MoleculesHost&              queriesHost,
+void preprocessRecursiveSmarts(SubstructTemplateConfig templateConfig,
+                               const MoleculesDevice&  targetsDevice,
+                               const MoleculesHost& /*queriesHost*/,
```
evasnow1992: `queriesHost` is unused, but `recursive_preprocessor.h:316` still takes it. May considering dropping it, which case also need to update other callers that still pass it.

evasnow1992 (approved):
Thank you for add the deduplication. Changes look good to me. One comment for cleaning up an unused parameter.

src/substruct/recursive_preprocessor.cu
scal444: Removed

PR #315, opened 2026-09-19
Enable most ruff checks

scal444:


evasnow1992 (approved):

PR #317, opened 2026-09-19
Enable targeted clang-tidy readability checks

scal444:
Enable a focused set of low-churn clang-tidy readability checks and apply their fixes to project sources.

This also scopes clang-tidy to `src/`, keeping `rdkit_extensions` aligned with upstream RDKit.

Validation:
- host clang-tidy pass across the in-scope translation units
- CUDA clang-tidy pass with no readability diagnostics
- Release GCC/NVCC build with `CMAKE_CUDA_ARCHITECTURES=89`
- full pre-push formatting and lint gate


scal444:
The one I care about most here is the statement braces, since the models love to do one line indented statements that look awful. #308

src/data_structures/flat_bit_vect.h
```diff
@@ -104,7 +104,7 @@ template <std::size_t NBits> struct FlatBitVectStorage {
 //! In-memory bitvect.
 template <std::size_t NBits> class FlatBitVect {
  public:
-  using StorageType                          = typename detail::FlatBitVectStorage<NBits>::StorageType;
+  using StorageType                          = detail::FlatBitVectStorage<NBits>::StorageType;
```
evasnow1992: Our somewhat stale `README.md` still lists Clang 15 as supported. Removing `typename` here causes Clang 15 to fail with `error: missing 'typename' prior to dependent type name`. Should we keep the `typename` for compatibility?

evasnow1992 (approved):
Changes look good to me. Just one comment regarding the removal of `typename`, which may cause a compilation error with Clang 15. Or we may want to update the readme file to recommend a different compiler/version?

src/data_structures/flat_bit_vect.h
scal444: Removed the check, since that's clang 15 not supporting that c++ 20 feature

